PR #896, opened 2017-08-22
CI: Build python and java artifacts in travis.ci

charliestrawn-wf:
@Workiva/messaging-pp this PR builds all of the libraries except for dart (need thrift on pub for that) and also runs the compiler tests.


aviary-wf:
## Raven
<details>

</details>

scripts/travis/deploy.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_BRANCH" = 'master' ] && [ "$TRAVIS_PULL_REQUEST" == 'false' ]; then
+    mvn deploy -P sign,build-extras --settings mvnsettings.xml
```
brianshannan-wf: Is this where we will need to publish to pypi in the future?

charliestrawn-wf: yep, this isn't being called yet. Will follow up with pushing the artifacts in the next PR

brianshannan-wf:
+1

PR #903, opened 2017-08-24
Memory fixes for GRC

brianshannan-wf:
I'm not sure how much of this is necessary as, in the browser, transports are usually very long lived. For the case of the nats transport (not in this repo, but extends classes in here) I believe one is instantiated at load time and used for the duration of the session.

Either way, I think the completer cleanup on messages needed to happen.

@Workiva/messaging-pp 
@Workiva/performance-reviewers I was told one of you could look over this to make sure everything looks fine

### DIFF
--- a/lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
+++ b/lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
@@ -17,7 +17,7 @@ part of frugal.src.frugal;
 /// with exponential backoff behavior and a capped number of retries. Its
 /// behavior can be customized by extending this class and overriding desired
 /// callbacks.
-class BaseFTransportMonitor extends FTransportMonitor {
+class BaseFTransportMonitor extends FTransportMonitor with Disposable {
   /// Default maximum reopen attempts.
   static const int DEFAULT_MAX_REOPEN_ATTEMPTS = 60;
 
@@ -44,6 +44,9 @@ class BaseFTransportMonitor extends FTransportMonitor {
     this._maxReopenAttempts = maxReopenAttempts;
     this._initialWait = initialWait;
     this._maxWait = maxWait;
+
+    manageStreamController(_onConnectController);
+    manageStreamController(_onDisconnectController);
   }
 
   /// Listen to connect events.
--- a/lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
+++ b/lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
@@ -17,7 +17,8 @@ part of frugal.src.frugal;
 /// with exponential backoff behavior and a capped number of retries. Its
 /// behavior can be customized by extending this class and overriding desired
 /// callbacks.
-class BaseFTransportMonitor extends FTransportMonitor with Disposable {
+@Deprecated('3.0.0')
+class BaseFTransportMonitor extends FTransportMonitor {
   /// Default maximum reopen attempts.
   static const int DEFAULT_MAX_REOPEN_ATTEMPTS = 60;
 
--- a/lib/dart/lib/src/frugal/transport/f_adapter_transport.dart
+++ b/lib/dart/lib/src/frugal/transport/f_adapter_transport.dart
@@ -33,6 +33,8 @@ class FAdapterTransport extends FAsyncTransport {
     transport.socket.onState.listen((state) {
       if (state == TSocketState.OPEN) _monitor?.signalOpen();
     });
+
+    manageDisposable(_framedTransport);
   }
 
   @override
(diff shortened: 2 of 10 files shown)


aviary-wf:
## Raven
<details>

</details>

lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
```diff
@@ -17,7 +17,7 @@ part of frugal.src.frugal;
 /// with exponential backoff behavior and a capped number of retries. Its
 /// behavior can be customized by extending this class and overriding desired
 /// callbacks.
-class BaseFTransportMonitor extends FTransportMonitor {
+class BaseFTransportMonitor extends FTransportMonitor with Disposable {
```
georgelesica-wf: Where will `dispose` be called from? If this thing can close itself then when it closes you may want to call `dispose` unless it can be re-opened.

lib/dart/lib/src/frugal/transport/f_adapter_transport.dart
```diff
@@ -41,17 +43,20 @@ class FAdapterTransport extends FAsyncTransport {
...
-      } catch (e) {
-        // Fatal error. Close the transport.
-        _adapterTransportLog.severe(
-            "FAsyncCallback had a fatal error ${e.toString()}." +
-                "Closing transport.");
-        close(e);
-      }
-    });
+
+    listenToStream(_framedTransport.onFrame, _handleFrame);
```
georgelesica-wf: Can this class be closed and then opened again? If so, this line will cause a problem because it will build up more and more subscriptions. If not, then you'll need to call `dispose` on close.

lib/dart/lib/src/frugal/transport/t_framed_transport.dart
```diff
@@ -15,7 +15,7 @@ part of frugal.src.frugal;
 
 /// A framed implementation of [TTransport]. Has stream for consuming
 /// entire frames. Disallows direct reads.
-class _TFramedTransport extends TTransport {
+class _TFramedTransport extends TTransport with Disposable {
```
georgelesica-wf: Where will this be disposed from? You may want to document that new requirement if this or any sub-classes end up being exported. Same with the `FTransport` class above.

lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
brianshannan-wf: Yeah most of our stuff is able to be re-opened after it is closed

lib/dart/lib/src/frugal/transport/f_adapter_transport.dart
brianshannan-wf: So this needs to be manually closed in `close()`

lib/dart/lib/src/frugal/transport/t_framed_transport.dart
brianshannan-wf: As I said in the description, I'm not entirely sure this needs to be disposable, as even though it has things like stream controllers, I think the transports exist for the entire lifetime of a browser session. @evanweible-wf might know better

lib/dart/lib/src/frugal/transport/f_adapter_transport.dart
georgelesica-wf: Yeah, probably, or you can call `dispose` in `close`, but then anything that is managed will have to be re-instantiated (or re-opened, etc) in the `open` method. You're probably better off doing it manually for clarity.

lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
georgelesica-wf: You may want to document to your consumers that it now needs to be disposed when they're finished with it then.

evanweible-wf:
Sorry for the delay, I wanted to pull messaging-sdk and frugal down to look into this further.

Brian is right that the underlying transports (nats messaging client, vessel messaging client) exist for the entire lifetime of a browser session, but those clients expose APIs for creating scope and service providers:

- `newHttpClient()` _*http can effectively be ignored as there is no persistent transport involved_
- `NatsMessagingClient.newClient()`
- `NatsMessagingClient.newPubSubProvider()`
- `VesselMessagingClient.newPubSubProvider()`

All of these providers/clients are created by and will need to be managed by consumers. The easiest way for this to work is if they are all disposable – because then consumers can simply call `manageDisposable()` on them to make sure they get disposed when their module or service disposes.

The other side of this is that all of these providers are tied directly to an underlying transport, which itself could be disposed. In that case, we would probably want all of the providers that were using said transport to dispose automatically since they would no longer be valid. Again, like you said, these underlying transports will exist for the entirety of the browser session _in practice_, but ensuring that everything is managed and gets cleaned up correctly will be important for testing, especially performance tests.

Most of this responsibility will fall to messaging-sdk, but there are some additional things that can be done here:

- `FTransport` – the `monitor` setter creates a new instance of `MonitorRunner()`, which is scoped to the `FTransport` instance upon which it was set. This `MonitorRunner` should be managed via `manageDisposable()` so that it stops listening when its transport disposes. (This would also require that `MonitorRunner` extend `Disposable`.)

- `FAsyncTransport` – both `oneway()` and `request()` kick off async actions that should probably be guarded with an `awaitBeforeDispose()` to ensure that the transport doesn't dispose until those async actions are complete.

- `FAdapterTransport` – the constructor listens to two streams on the transport's socket that should be managed via `listenToStream()`

- Anything that can be opened/closed multiple times and is also `Disposable` should have two things done:
  - The difference between the two (`close()` being temporary, can be restored via `open()`, whereas `dispose()` is final) should be clearly explained in the doc comments.
  - `onDispose()` should be overridden to call and wait for `close()`

stevenosborne-wf:
+1

stevenosborne-wf:
@evanweible-wf are these changes sufficient?

tylerrinnan-wf:
lgtm

lib/dart/lib/src/frugal/transport/base_f_transport_monitor.dart
```diff
@@ -17,6 +17,7 @@ part of frugal.src.frugal;
 /// with exponential backoff behavior and a capped number of retries. Its
 /// behavior can be customized by extending this class and overriding desired
 /// callbacks.
+@Deprecated('3.0.0')
```
evanweible-wf: Not a big deal for this PR, but for future reference we're advocating the following pattern for deprecations:
https://dev.workiva.net/docs/platform/application-frameworks/front-end-dev-policies/wdesk-versioning/deprecation-strategies

PR #905, opened 2017-08-30
change log name in dart generated code

brianshannan-wf:
This naming of our logger was causing a conflict when an IDL as `log()` as a service method. This added a frugal prefix, which I intended to be viewed as a "reserved" prefix, implying users should not use this prefix for the names in their IDL

@Workiva/messaging-pp @markerickson-wf


aviary-wf:
## Raven
<details>

</details>

markerickson-wf:
lgtm - thanks

stevenosborne-wf:
+1

PR #913, opened 2017-09-13
Switch frugal binary and library to use glide instead of godep

charliestrawn-wf:
This seems to be working, need to fix the travis python permissions issues


aviary-wf:
## Raven
<details>

</details>

scripts/skynet/run_cross_skynet.sh
```diff
@@ -25,6 +25,8 @@ frugal --gen py:asyncio -r --out='test/integration/python/asyncio/gen_py_asyncio
 frugal --gen py -r -out='test/integration/python/tornado/gen-py' test/integration/frugalTest.frugal
 frugal --gen dart:use_enums=true -r --out='test/integration/dart/gen-dart' test/integration/frugalTest.frugal
 
+go get github.com/Sirupsen/logrus
```
brianshannan-wf: Why do we need to explicitly go get this?

charliestrawn-wf: The tests use it, but they are run from a test/integration dir. Frugal doesn't use it, so when it walks the gopath it never finds it. I can look at just using the go library logger for integration tests going forward.

brianshannan-wf:
@aviary-wf

lib/go/glide.yaml
```diff
@@ -0,0 +1,46 @@
+package: github.com/Workiva/frugal/lib/go
+import:
+- package: git.apache.org/thrift.git
+  version: 0d9b713b173f35ce02552b2f4372899440a99b25
```
brianshannan-wf: Should we pin this to an actual tag release? Same question for others below

scripts/skynet/run_cross_skynet.sh
brianshannan-wf: Doesn't that directory have it's own glide.yaml?

rmconsole-wf:
@charliestrawn-wf This pull request has merge conflicts, please resolve.

brianshannan-wf:
Closing in favour of https://github.com/Workiva/frugal/pull/976

PR #914, opened 2017-09-14
don't show dependencies multiple times in the html index

brianshannan-wf:
If a dependency was included multiple times via a transitive dependency, like `b` depends on `a` and `c` depends on both `b` and `a`, then `a` shouldn't be shown multiple times in the index generated by the html generator.

@Workiva/messaging-pp @markcampanelli-wf


aviary2-wf:
## Raven
<details>

</details>

brianshannan-wf:
travis is currently failing for an unrelated problem

charliestrawn-wf:
+1

tylerrinnan-wf:
lgtm

PR #916, opened 2017-09-18
Python serialize

johnlockwood-wf:
Add frugal serialize functions which cope with unicode string values for structure properties. They are not handled by Thrift serialize functions with frugal generated code.

@Workiva/messaging-pp


aviary2-wf:
## Raven
<details>

</details>

PR #927, opened 2017-10-19
[WIP] Python: Add 'additional_headers' param to FHTTPTransport

tylerwilding-wk:

PR #922, opened 2017-10-06
java: Add FHttpTransportHeaders.getRequestHeaders(FContext)

brettkail-wk:
Add `getRequestHeaders(FContext)` to `FHttpTransportHeaders`.  Implement it as a default method that calls `getRequestHeaders()` so that behavior of existing implementations is unchanged.

Fixes #920

Issue #920, opened 2017-10-05
[Feature Request] Add http headers on a per-request basis


macleodbroad-wf:
## Problem

Currently it is not possible to add http headers on a per-request basis.

## Current Work-Around (Inefficient)

Because some http header values change on every request you'd need to completely re-build the client before every request:
```java
FServiceProvider serviceProvider = (new HttpMessaging.ClientBuilder(null, serviceDescriptor)).withRequestHeaders(requestHeaders).build();
```

## Potential Solutions

@brettkail-wk and I did some brainstorming on potential solutions. The solutions below are ordered from Most Preferred to Least Preferred.

### Add an attribute to the FContext

`FContext.addAttribute("com.workiva.frugal.http.headers")`, and FHttpTransport.addHttpHeader(FContext, String key, String value)

This is by far the simplest, with none of the drawbacks of the other approaches and offers maximum flexibility to consumers.

### Whitelisted headers picked from FContext Request Headers and injected into the Http Headers

1. Whitelist of header keys passed into the `HttpMessaging.ClientBuilder.withRequestHeadersFromContext`
2. `makeRequest` cycles through and picks the whitelisted headers out of the `FContext` passed in on the request, and adds them as HTTP Headers
```
requestHeaderKeys = ['X-Cloud-Trace-Context', 'Some-Other-Header'... ];
FServiceProvider serviceProvider = (new HttpMessaging.ClientBuilder(null, serviceDescriptor))
                                   .withRequestHeadersFromContext(requestHeaderKeys)
                                   .build();
...
    private byte[] makeRequest(FContext context, byte[] requestPayload) throws TTransportException {
        // Encode request payload
        String encoded = Base64.encodeBase64String(requestPayload);
        StringEntity requestEntity = new StringEntity(encoded, ContentType.create("application/x-frugal", "utf-8"));
        // Set headers and payload
        HttpPost request = new HttpPost(url);
        // add user supplied headers first, from requestHeaderKeysFromContext list
        if (requestHeaderKeysFromContext != null) {
            for (String key : requestHeaderKeysFromContext) {
				String value = fContext.getRequestHeader(key);
                if (key != null && value != null) {
                    request.setHeader(key, value);
                }
            }
        ...
```
This is a simple approach, but has obvious draw-backs.

### Override getRequestHeaders in FHttpTransportHeaders to accept a context

See https://github.com/brettkail-wk/frugal/commit/c593098367a23600980e617e10fc13282611da77

[...]

PR #928, opened 2017-10-19
Python: Add param to allow for additional request headers to FHTTPTransport #927

tylerwilding-wk:
added `get_request_header` function to tornado and asyncio's `FHttpTransport`.  By default it will not add any additional headers onto the request so existing implementation is not effected, but can be overridden by a consumer to append additional headers.

Addresses #920 and based off changes in #922

### DIFF
--- a/lib/python/frugal/aio/transport/http_transport.py
+++ b/lib/python/frugal/aio/transport/http_transport.py
@@ -29,7 +29,7 @@ class FHttpTransport(FTransportBase):
     This allows messages of arbitrary sizes to be sent and received.
     """
     def __init__(self, url, request_capacity=0, response_capacity=0,
-                 request_header_funcs=[]):
+                 request_header_func=None):
         """
         Create an HTTP transport.
 
@@ -39,18 +39,18 @@ class FHttpTransport(FTransportBase):
                               request. Set to 0 for no size restrictions.
             response_capacity: The maximum size allowed to be read in a
                                response. Set to 0 for no size restrictions
-            request_header_funcs: An optional list of functions that
-                                  return a dictionary of request headers
-                                  to be added to the request. Set to
-                                  an empty list by default.
+            request_header_func: An optional lambda function that will
+                                 return a dictionary of request headers when
+                                 called to be added to the request. Set to
+                                 None by default.
         """
         super().__init__(request_capacity)
         self._url = url
         self._headers = {}
 
         # Add user-supplied headers first to avoid removing the native headers
-        for func in request_header_funcs:
-            for header, value in func.items():
+        if request_header_func is not None:
+            for header, value in request_header_func().items():
                 if header is not None and value is not None:
                     self._headers[header] = value
 
--- a/lib/python/frugal/aio/transport/http_transport.py
+++ b/lib/python/frugal/aio/transport/http_transport.py
@@ -28,7 +28,8 @@ class FHttpTransport(FTransportBase):
     FHttpTransport is an FTransport that uses http as the underlying transport.
     This allows messages of arbitrary sizes to be sent and received.
     """
-    def __init__(self, url, request_capacity=0, response_capacity=0):
+    def __init__(self, url, request_capacity=0, response_capacity=0,
+                 additional_headers=None):
         """
         Create an HTTP transport.
 
@@ -38,6 +39,9 @@ class FHttpTransport(FTransportBase):
                               request. Set to 0 for no size restrictions.
             response_capacity: The maximum size allowed to be read in a
                                response. Set to 0 for no size restrictions
+            additional_headers: Additional headers to be appended onto
+                                the request. Set to None by default, expects
+                                a dict of key-value pairs
         """
         super().__init__(request_capacity)
         self._url = url
@@ -49,6 +53,10 @@ class FHttpTransport(FTransportBase):
         }
         if response_capacity > 0:
             self._headers['x-frugal-payload-limit'] = str(response_capacity)
+        # append additional provided headers
+        if additional_headers is not None:
+            for header, value in additional_headers.items():
+                self._headers[header] = str(value)
 
     def is_open(self):
         """Always returns True"""
--- a/lib/python/frugal/aio/transport/http_transport.py
+++ b/lib/python/frugal/aio/transport/http_transport.py
@@ -29,7 +29,7 @@ class FHttpTransport(FTransportBase):
     This allows messages of arbitrary sizes to be sent and received.
     """
     def __init__(self, url, request_capacity=0, response_capacity=0,
-                 request_header_func=None):
+                 get_request_headers=None):
         """
         Create an HTTP transport.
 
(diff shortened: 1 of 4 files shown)


aviary2-wf:
## Raven
<details>

</details>

brianshannan-wf:
@Workiva/messaging-pp

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -49,6 +53,10 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
         }
         if response_capacity > 0:
             self._headers['x-frugal-payload-limit'] = str(response_capacity)
+        # append additional provided headers
```
brianshannan-wf: Should set these before the headers set by frugal so the native frugal headers can't get over written

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -49,6 +53,10 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
...
+        if additional_headers is not None:
+            for header, value in additional_headers.items():
+                self._headers[header] = str(value)
```
brianshannan-wf: I don't think this cast should be here, as if a non-string object is provided it most likely won't work as intended, and will have a harder debugging path to find out what went wrong than if an exception was thrown somewhere

lib/python/frugal/tests/aio/transport/test_http_transport.py
```diff
@@ -33,8 +33,17 @@ def setUp(self):
             request_capacity=self.request_capacity,
             response_capacity=self.response_capacity
         )
+        self.transport_headers = FHttpTransport(
```
brianshannan-wf: maybe `transport_with_headers`? To me, `transport_headers` implies a dictionary like object, not a transport

brianshannan-wf: Alternatively, if this is only used in one test it might make more sense to belong solely in that test

tylerwilding-wk:
fyi - @macleodbroad-wf

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -107,6 +101,15 @@ def is_open(self):
 
         return TMemoryBuffer(decoded[4:])
 
+    async def get_request_headers(self, fcontext):
```
brianshannan-wf: I think this function should be passed into the transport instead of requiring the transport to be inherited from and this method overridden.

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -118,12 +121,26 @@ def is_open(self):
         Throws:
             TTransportException if the request timed out.
         """
+
+        request_headers = await self.get_request_headers(fcontext=context)
```
brianshannan-wf: Nit: don't pass as a kwarg as you have the function taking a position argument

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -28,7 +28,8 @@ class FHttpTransport(FTransportBase):
     FHttpTransport is an FTransport that uses http as the underlying transport.
     This allows messages of arbitrary sizes to be sent and received.
     """
-    def __init__(self, url, request_capacity=0, response_capacity=0):
+    def __init__(self, url, request_capacity=0, response_capacity=0,
+                 request_header_funcs=[]):
```
tylertreat: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -38,15 +39,25 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
...
+                                  to be added to the request. Set to
+                                  an empty list by default.
         """
         super().__init__(request_capacity)
         self._url = url
+        self._headers = {}
+
+        # Add user-supplied headers first to avoid removing the native headers
+        for func in request_header_funcs:
+            for header, value in func.items():
```
tylertreat: Shouldn't you be calling the function?

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -38,11 +39,27 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
...
+                                  to be added to the request. Set to
+                                  an empty list by default.
         """
         super().__init__(request_capacity)
         self._url = url
+        self._headers = {}
 
-        self.response_capacity = response_capacity
+        # Add user-supplied headers first to avoid removing the native headers
+        for func in request_header_funcs:
```
brianshannan-wf: What's the use case for needing multiple functions? It seems like one function would work, which is what our java implementation currently does

lib/python/frugal/aio/transport/http_transport.py
tylerwilding-wk: oh wow, thanks I would have never guessed that.

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -38,15 +39,25 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
...
+                                 called to be added to the request. Set to
+                                 None by default.
         """
         super().__init__(request_capacity)
         self._url = url
+        self._headers = {}
+
+        # Add user-supplied headers first to avoid removing the native headers
+        if request_header_func is not None:
+            for header, value in request_header_func().items():
```
brianshannan-wf: Just to clarify @macleodbroad-wf as you're tagged on this pr, having a function that adds the same headers to all requests will satisfy your use case? As I remember you needed the ability to add dynamically created headers to requests in java

macleodbroad-wf: I'm sorry. I'm not sure I understand.

The use-case is that a single instance of a client will have a unique value for the `X-Cloud-Trace-Context` header on each request.

tylertreat: IMO this should match what was done in Java to allow providing a closure that takes in an `FContext`.

macleodbroad-wf: I agree with that.

brianshannan-wf: Yep, so this won't accomplish that goal, you'll need to copy the `self._headers` and call the user provided function in each request, not in `__init__()`

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -38,9 +39,13 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
                               request. Set to 0 for no size restrictions.
             response_capacity: The maximum size allowed to be read in a
                                response. Set to 0 for no size restrictions
+            get_request_headers: An optional function that accepts an FContext.
+                                 Should return a dictionary of additional
+                                 request headers to be appended onto the request
         """
         super().__init__(request_capacity)
         self._url = url
+        self.get_request_headers = get_request_headers
```
brianshannan-wf: Make this `self._get_request_headers` so it's not exported as part of the public api

lib/python/frugal/tornado/transport/http_transport.py
```diff
@@ -37,11 +38,15 @@ def __init__(self, url, request_capacity=0, response_capacity=0):
...
                                response. Set to 0 for no size restrictions.
+            get_request_headers: An optional function that accepts an FContext.
+                                 Should return a dictionary of additional
+                                 request headers to be appended onto the request
         """
         super(FHttpTransport, self).__init__(
             request_size_limit=request_capacity)
         self._url = url
         self._http = AsyncHTTPClient()
+        self.get_request_headers = get_request_headers
```
brianshannan-wf: same comment

brianshannan-wf:
Looks good after some minor things are fixed

brianshannan-wf:
+1

lib/python/frugal/tornado/transport/http_transport.py
```diff
@@ -84,12 +89,20 @@ def request(self, context, payload):
         """
         Write the current buffer and return the response.
         """
+        # construct headers for request
+        request_headers = {}
+        if self._get_request_headers is not None:
+            request_headers = self._get_request_headers(context)
+        # apply the default headers so their values cannot be modified
+        for header, value in self._headers.iteritems():
+            request_headers[header] = value
```
macleodbroad-wf: I think you can replace this for loop with
```python
request_headers.update(self._headers)
```

macleodbroad-wf:
+1

brianshannan-wf:
+1

macleodbroad-wf:
+1

charliestrawn-wf:
+1 - thanks for the contribution!

PR #930, opened 2017-10-24
Move multidict requirement to asyncio extras

johnlockwood-wf:
WIP until I can get the skynet tests to pass

Upgrade aiohttp. The latest aiohttp doesn't depend on multidict anymore, so our test that uses it doesn't need it anymore.
The multidict dep was also in the main requirements, but was only needed in the asyncio extras.

@Workiva/messaging-pp


aviary2-wf:
## Raven
<details>

</details>

brianshannan-wf:
This pr needs to be made against the develop branch

compiler/generator/python/generator.go
```diff
@@ -1270,7 +1270,7 @@ func (g *Generator) generateProcessorFunction(method *parser.Method) string {
 		contents += tabtabtab + "with self._lock:\n"
 		contents += tabtabtabtab + fmt.Sprintf("_write_application_exception(ctx, oprot, \"%s\", ex_code=TApplicationExceptionType.INTERNAL_ERROR, message=e.message)\n", methodLower)
 	}
-	contents += tabtabtab + "raise\n"
+	contents += tabtabtab + "raise e\n"
```
aldenpeterson-wf: can you clarify why you are doing this?

This will lose the exception stack history.

charliestrawn-wf:
@johnlockwood-wf did you mean to add the commits for switching to dep to this PR? The massive diff makes the original change impossible to review. Also we have a PR open for switching from godep to glide. I'm all for using dep but that was shot down as Workiva's dependency tracking tools don't support it yet.

johnlockwood-wf:
@charliekump-wf I meant to make the crossrunner build following a change and it wouldn't build using the godep workspaces setup.

johnlockwood-wf:
closing. will extract the cross test enhancements after the switch from godeps happens.

PR #934, opened 2017-10-30
dart: add getRequestHeaders param to FHttpTransport

tylerwilding-wk:
dart equivalent of recent python changes - https://github.com/Workiva/frugal/pull/928

Allows consumer to pass in a function that accepts an FContext using the `getRequestHeaders` parameter.  This function should return additional request headers to be appended on a per-request basis.

@Workiva/messaging-pp

### DIFF
--- a/lib/dart/lib/src/frugal/transport/f_http_transport.dart
+++ b/lib/dart/lib/src/frugal/transport/f_http_transport.dart
@@ -13,6 +13,9 @@
 
 part of frugal.src.frugal;
 
+/// typedef for passed in function that generates headers given an [FContext]
+typedef Map<String, String> GetHeadersWithContext(FContext ctx);
+
 /// An [FTransport] that makes frugal requests via HTTP.
 class FHttpTransport extends FTransport {
   /// HTTP status code for an unauthorized reqeuest.
@@ -37,7 +40,7 @@ class FHttpTransport extends FTransport {
 
   /// Function that accepts an FContext that should return a Map<String, String>
   /// of headers to be added to every request
-  var _getRequestHeaders;
+  final GetHeadersWithContext _getRequestHeaders;
 
   /// Create an [FHttpTransport] instance with the given w_transport [Client],
   /// uri, and optional size restrictions, and headers.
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/dart/lib/src/frugal/transport/f_http_transport.dart
```diff
@@ -35,6 +35,10 @@ class FHttpTransport extends FTransport {
 
   Map<String, String> _headers;
 
+  /// Function that accepts an FContext that should return a Map<String, String>
+  /// of headers to be added to every request
+  var _getRequestHeaders;
```
brianshannan-wf: A couple of changes would be good here:
1. Make this final
2. Give this a proper type (instead of var)

lib/dart/lib/src/frugal/transport/f_http_transport.dart
```diff
@@ -44,18 +48,28 @@ class FHttpTransport extends FTransport {
...
+  ///
+  /// Additionally, a function that accepts an FContext can be passed in
+  /// that should return additional headers to be appended to each request
+  /// using the getRequestHeaders param.
   FHttpTransport(this.client, this.uri,
       {int requestSizeLimit: 0,
       this.responseSizeLimit: 0,
-      Map<String, String> additionalHeaders})
+      Map<String, String> additionalHeaders,
+      var getRequestHeaders: null})
```
brianshannan-wf: Should do `...<type> getRequestHeaders: null}) : this._getRequestHeaders = getRequestHeaders`

brianshannan-wf: Actually doing the default to a no-op function here would be even better

lib/dart/test/transport/f_http_transport_test.dart
```diff
@@ -111,6 +117,90 @@ void main() {
...
+          return new MockResponse.badRequest();
+        }
+      });
+
+      var response = await transportWithContext.request(
+          newContext, transportRequest) as TMemoryTransport;
+      expect(response.buffer, transportResponse);
+    });
+
+    test(
```
brianshannan-wf: Do these tests need to exist for `transportWithContext`? It doesn't seem like they're testing anything related to adding headers to the http request

tylerwilding-wk: Seems like they aren't, at one point I had to add them to get the actually useful test passing, but I suppose it was just a red herring.  Removed

lib/dart/lib/src/frugal/transport/f_http_transport.dart
tylerwilding-wk: I like the idea of defaulting to a no-op function (just returns an empty Map) as that could clean up the logic later on.  But I'm not sure how to do that in Dart, default values must be constants and methods cannot be declared as constants?

brianshannan-wf:
+1

charliestrawn-wf:
lgtm

PR #936, opened 2017-11-01
Move multidict requirement to asyncio extras

johnlockwood-wf:
This isolates the issue fix from https://github.com/Workiva/frugal/pull/930. I'll wait on adding the changes to the skynet cross tests until the switch from Godeps is complete. One of the cross test changes is to make it output a csv table with all the variables and outcome of each client/server combination test. It makes it so you can quickly find the factors associated with failures and would have made finding this fix a lot faster.

Move multidict to its proper place in asyncio extras

Fix issue where integration test couldn't handle some exceptions

Work around bug in aiohttp

Pull requests should be made against the 'develop' branch, not master.

@Workiva/messaging-pp

### DIFF
--- a/lib/python/Makefile
+++ b/lib/python/Makefile
@@ -4,6 +4,8 @@ PIP := pip
 EXCLUDE_GAE := --exclude-dir=frugal/gae --exclude-dir=frugal/tests/gae
 EXCLUDE_TORNADO := --exclude-dir=frugal/tornado --exclude-dir=frugal/tests/tornado
 EXCLUDE_AIO := --exclude-dir=frugal/aio --exclude-dir=frugal/tests/aio
+COVER := --with-coverage --cover-package=frugal
+LEVEL := --logging-level=ERROR
 
 all: deps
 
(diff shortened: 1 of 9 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/python/Makefile
```diff
@@ -43,5 +46,9 @@ xunit-py2:
 xunit-py3:
 	nosetests --exclude-dir=frugal/tornado --exclude-dir=frugal/tests/tornado --exclude-dir=frugal/gae --exclude-dir=frugal/tests/gae --logging-level=ERROR --with-coverage --cover-package=frugal --with-xunit --xunit-file=unit_tests_py3.xml
```
brianshannan-wf: If we're going to add excludes we should probably use them in these make targets as well

scripts/skynet/cross/setup_python.sh
```diff
@@ -11,4 +11,5 @@ else
 fi
 
 pip install -e ".[tornado]"
+python3.5 /usr/bin/pip3 install Cython==0.27.2
```
brianshannan-wf: Does this need to be here? I'd prefer to keep the environment as close to a fresh pip install as possible

lib/python/Makefile
charliestrawn-wf: might also be nice to add a constant for these `--logging-level=ERROR --with-coverage --cover-package=frugal` since they are used in most of the targets

scripts/skynet/cross/setup_python.sh
johnlockwood-wf: There was an issue installing multidict without Cython. Installing would skip building the c extension and some part of the process expects the .c file to be there so it fails.

johnlockwood-wf: From what I read, pure python multidict is an order of magnitude slower than a regular dict and the extension makes it performant enough

charliestrawn-wf:
lgtm

brianshannan-wf:
+1

PR #942, opened 2017-11-16
[WIP] Inner provider middleware

brianshannan-wf:
A use case has come up where potentially sensitive information should be stripped from an FContext on a publish call, as that can potentially get back to clients. This could be done easily in an enforced middleware, but would need a slightly different pattern than is currently supported.

Currently we only allow enforced middleware to be the outermost middleware, while we would want this middleware to be at the innermost level, so other middleware couldn't add it back.

This seemed like the best solution, as it accomplishes the goal in an easy to consume way and shouldn't break anything.

TODO if this gets approval add to other languages

@Workiva/messaging-pp @tylertreat @brettkail-wk

### DIFF
--- a/lib/go/provider.go
+++ b/lib/go/provider.go
@@ -38,6 +38,7 @@ func NewFScopeProvider(pub FPublisherTransportFactory, sub FSubscriberTransportF
 }
 
 // NewFScopeProvider2
+// TODO better name
 func NewFScopeProvider2(pub FPublisherTransportFactory, sub FSubscriberTransportFactory,
 	prot *FProtocolFactory, outerMiddleware []ServiceMiddleware, innerMiddleware []ServiceMiddleware) *FScopeProvider {
 	return &FScopeProvider{
@@ -89,7 +90,8 @@ func (p *FScopeProvider) GetInnerMiddleware() []ServiceMiddleware {
 type FServiceProvider struct {
 	transport       FTransport
 	protocolFactory *FProtocolFactory
-	middleware      []ServiceMiddleware
+	outerMiddleware      []ServiceMiddleware
+	innerMiddleware []ServiceMiddleware
 }
 
 // NewFServiceProvider creates a new FServiceProvider containing the given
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/go/provider.go
```diff
@@ -86,8 +128,21 @@ func (f *FServiceProvider) GetProtocolFactory() *FProtocolFactory {
 }
 
 // GetMiddleware returns the ServiceMiddleware stored on this FServiceProvider.
-func (f *FServiceProvider) GetMiddleware() []ServiceMiddleware {
-	middleware := make([]ServiceMiddleware, len(f.middleware))
-	copy(middleware, f.middleware)
+// DEPRECATED: replaced by GetOuterMiddleware() for more specificity
+func (p *FServiceProvider) GetMiddleware() []ServiceMiddleware {
+	return p.GetMiddleware()
```
brettkail-wk: Should be `GetOuterMiddleware`.

brianshannan-wf: good catch

lib/go/provider.go
```diff
@@ -31,7 +32,21 @@ func NewFScopeProvider(pub FPublisherTransportFactory, sub FSubscriberTransportF
...
-		middleware:                 middleware,
+		outerMiddleware:            middleware,
+		innerMiddleware:            []ServiceMiddleware{},
+	}
+}
+
+// NewFScopeProvider2
+// TODO better name
+func NewFScopeProvider2(pub FPublisherTransportFactory, sub FSubscriberTransportFactory,
+	prot *FProtocolFactory, outerMiddleware []ServiceMiddleware, innerMiddleware []ServiceMiddleware) *FScopeProvider {
```
tylertreat: Do you need this constructor? Can you just expose setters?

lib/go/provider.go
```diff
@@ -50,9 +65,22 @@ func (p *FScopeProvider) NewSubscriber() (FSubscriberTransport, *FProtocolFactor
 }
 
 // GetMiddleware returns the ServiceMiddleware stored on this FScopeProvider.
+// DEPRECATED: replaced by GetOuterMiddleware() for more specificity
 func (p *FScopeProvider) GetMiddleware() []ServiceMiddleware {
-	middleware := make([]ServiceMiddleware, len(p.middleware))
-	copy(middleware, p.middleware)
+	return p.GetMiddleware()
```
tylertreat: This looks like a stack overflow?

lib/go/provider.go
```diff
@@ -50,9 +65,22 @@ func (p *FScopeProvider) NewSubscriber() (FSubscriberTransport, *FProtocolFactor
...
+}
+
+// GetOuterMiddleware
+func (p *FScopeProvider) GetOuterMiddleware() []ServiceMiddleware {
+	middleware := make([]ServiceMiddleware, len(p.outerMiddleware))
+	copy(middleware, p.outerMiddleware)
+	return middleware
+}
+
+// GetInnerMiddleware
```
tylertreat: Would be good to doc these properly.

lib/go/provider.go
```diff
@@ -62,7 +90,8 @@ func (p *FScopeProvider) GetMiddleware() []ServiceMiddleware {
 type FServiceProvider struct {
 	transport       FTransport
 	protocolFactory *FProtocolFactory
-	middleware      []ServiceMiddleware
+	outerMiddleware      []ServiceMiddleware
```
tylertreat: formatting

brianshannan-wf:
Closing as this actually isn't needed at the moment

PR #943, opened 2017-11-22
Java: Use ExecutorService.execute instead of submit if Future is unused

brettkail-wk:
If an exception is thrown from a task passed to `ExecutorService.submit`, the exception is captured in the returned `Future`.  If the `Future` is ignored, the exception message+stack is lost, which is problematic if `Error` is thrown from a request processor.  Use `ExecutorService.execute` in those cases, which allows the default exception handler to print the exception message+stack.

@Workiva/messaging-pp


aviary-wf:
## Raven
<details>

</details>

brianshannan-wf:
+1

kevinsookocheff-wf:
+1

PR #944, opened 2017-11-30
dart: Add deprecated comment to abstract classes

brettkail-wk:
Currently, the message from the `(deprecated="...")` annotation is not propagated to the abstract base class ("interface"), so it is not easily visible to clients.  Add `/// Deprecated: ...`.

Additionally, use the same pattern in the client class rather than `@Deprecated("...")` since that syntax sets an "expires" member that is used to describe when the member will be removed rather than being used as a general string to describe the deprecation, which is how it is used elsewhere.

@Workiva/messaging-pp


aviary2-wf:
## Raven
<details>

</details>

brianshannan-wf:
+1

charliestrawn-wf:
+1

PR #948, opened 2017-12-13
Use expanded docs from doc portal.

kevinsookocheff-wf:
Use expanded docs from Workiva's internal doc portal.


aviary-wf:
## Raven
<details>

</details>

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+  used to send OAuth2 tokens, distributed tracing metadata, and user-context
+  information, avoiding the need to include it everywhere in your IDL and
+  handler logic. Correlation IDs are also built into the request context for
+  tying logs, traces, and anytics together.
 
 - Middleware: client- and server- side middleware is supported for RPC and
   pub/sub APIs. This allows you to implement interceptor logic around handler
   functions, e.g. for authentication, logging, or retry policies.
+
+# Frugal Architecture
```
brianshannan-wf: Should we add a table of contents or break this out into a docs directory? There's a lot of info here

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+    ...
+    2: Time time
+    ...
+}
+```
+
+If you are designing an interface with a semantic type implemented in
+terms of a base type, but the semantic type is particularly significant or
+widely used it may be a good candidate for a typedef. Typedef types are
+self-documenting and, in statically typed languages, ensure that the
```
brianshannan-wf: nit: This isn't completely correct as java is statically typed and doesn't use typedefs. Maybe "...and, in languages that support them, ensure..."?

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+
+In most situations default requiredness is a good compromise, because it
+is always serialized but need not be present during deserialization. If
+you would like the flexibility to decide whether a field is serialized or
+not, choose optional requiredness. Required requiredness fields create
+runtime errors when not found during deserialization and cannot evolve,
+they should be avoided unless you want to permanently enforce a field’s
+presence.
+
+## Type Evolution
```
brianshannan-wf: Maybe put in a bit about frugal's backward compatibility audit feature?

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+```
+
+This variable is then passed to publish and subscribe calls:
+
+```go
+var (
+    event         = &event.Event{ID: 42, Message: "hello, world!"}
+    membershipID  = "ALKJ123LK90"
+)
+publisher.PublishEventCreated(frugal.NewFContext(""), event, membershipID)
```
brianshannan-wf: nit: `membershipID` should be the second argument, not `event`

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+
+subscriber.SubscribeEventCreated(user, func(ctx *frugal.FContext, e *event.Event) {
+    fmt.Printf("Received event for %s: %s\n", membershipID, e.Message)
+})
+```
+
+All messages being routed through the Messaging Frontend must be prefixed
```
brianshannan-wf: I don't think the next two paragraphs should be here as they are Workiva specifc

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+}
+
+scope Events {
+    EventCreated: Event
+}
+```
+
+## Prefixes
+
+By default, Frugal publishes messages on the topic `<scope>.<operation>`.
```
brianshannan-wf: technically, `frugal.` is prefixed to all topics by the publisher transports, so it would be `frugal.<scope>.<operation>`. Not sure how valuable that is to mention

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+executed, and unregistered.
+
+The server implementation is used on the server side, which is handling RPCs.
+It does not actually register FAsyncCallbacks but rather has an FProcessor
+registered with it. When a message is received, it's buffered and passed to
+the FProcessor to be handled.
+
+## FTransport
+
+FTransport is Frugal's equivalent of Thrift's TTransport. FTransport extends
```
brianshannan-wf: The "FTransport extends and adds additional methods" bit isn't very true as of 2.x. It's more like a completely redesigned TTransport

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+typically implemented by wrapping a TFramedTransport.
+
+Most Frugal language libraries include an FMuxTransport implementation, which
+uses a worker pool to handle messages in parallel.
+
+## FTransportFactory
+
+FTransportFactory produces FTransports by wrapping a provided TTransport.
+
+## FTransportMonitor
```
brianshannan-wf: These are either deprecated or deleted. Either way, they aren't needed and shouldn't be used

documentation/README.md
```diff
@@ -71,10 +71,967 @@ provides.
...
+
+## FProtocol
+
+FProtocol is Frugal's equivalent of Thrift's TProtocol. It defines the
+serialization protocol used for messages, such as JSON, binary, etc. FProtocol
+actually extends TProtocol and adds support for serializing FContext. In
+practice, FProtocol simply wraps a TProtocol and uses Thrift's built-in
+serialization. FContext is encoded before the TProtocol serialization of the
+message using a simple binary protocol. See the
+[protocol documentation](protocol.md) for more details.
```
brianshannan-wf: This should be updated if we keep everything in one markdown file

charliestrawn-wf:
closing as this is quite stale

PR #951, opened 2017-12-21
Acquire lock to read from transport

johnlockwood-wf:
Executing Close on an Adapter races the readLoop goroutine to read a frame from the transport.

@stevenosborne-wf 
@patkujawa-wf 
@ericklaus-wf

### DIFF
--- a/lib/go/adapter_transport.go
+++ b/lib/go/adapter_transport.go
@@ -117,6 +117,8 @@ func (f *fAdapterTransport) readLoop() {
 }
 
 func (f *fAdapterTransport) readFrame(framedTransport *TFramedTransport) ([]byte, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	_, err := framedTransport.Read([]byte{})
 	if err != nil {
 		return nil, err


aviary2-wf:
## Raven
<details>

</details>

tylertreat:
I think this will deadlock you if you're caught in a situation where there is no more data coming from the transport.

johnlockwood-wf:
Because it is going to block on the read?

patkujawa-wf:
Is there no timeout on the read? What happens when close is called while reading?

johnlockwood-wf:
@tylertreat Would you explain how this could cause a deadlock?

brianshannan-wf:
If you put a lock around the read call, it will hold the lock until the transport reads something. If nothing ever arrives for the transport to read, close will never be able to acquire the lock

patkujawa-wf:
Any update, guys?

stevenosborne-wf:
The issue ultimately comes from the fact that we are using a synchronous thrift transport in frugal's asynchronous environment. Frugal transports constantly read from the server, unlike in thrift where the read only happens after a particular request is issued. A lock around the Read call will block many interactions with the transport (as Brian mentioned above). Closing this PR.

johnlockwood-wf:
Can't we solve it with more indirection? If Close is called, instead of directly closing the transport, it sends to a close channel and the goroutine selects on both that channel and reading the frame.

brianshannan-wf:
But reading frames off the transport doesn't happen via a channel, it's a blocking call, so you can't select over that and something else

PR #952, opened 2017-12-28
Move vendored files to be rooted in vendor directory instead of vendor/src

brandonbodnar-wk:
Based on the Go command documentation, vendored code should be rooted to the "vendor" directory itself and not placed in a src subfolder.
See example for crash/bang at https://golang.org/cmd/go/#hdr-Vendor_Directories

Confirmed this as an issue by running running `go test` locally on the lib/go directory. Compiling failed due to an `go test` using the locally installed version of thift/lib/go in local $GOPATH, instead of the vendored thrift/lib/go in the vendor directory.

After this change, was able to run `go test` successfully as it now picks up the code in the vendor folder.


aviary-wf:
## Raven
<details>

</details>

PR #953, opened 2018-01-02
Java: Improve exception handling

brettkail-wk:
Three commits are intended to be reviewed independently, but all three improve Java exception handling:

1. Improve logging of undeclared `TException`.  See commit message for more details.
2. Document lack of `writeApplicationException` for `TException` in `FBaseProcessor`.  Noticed while trying to understand the exception handling of item 1 above.
3. Don't log RuntimeException twice for NATS.  As of commit 21758e33, RuntimeException are logged twice: once in `FBaseProcessor` and again by the Java uncaught exception handler.

@Workiva/messaging-pp

PR #954, opened 2018-01-04
Java: Improve exception handling

brettkail-wk:
(Updated #953 to fix build by keeping the same `TApplicationException` message.)

Three commits are intended to be reviewed independently, but all three improve Java exception handling:

1. Improve logging of undeclared `TException`.  See commit message for more details.
2. Document lack of `writeApplicationException` for `TException` in `FBaseProcessor`.  Noticed while trying to understand the exception handling of item 1 above.
3. Don't log RuntimeException twice for NATS.  As of commit 21758e33, RuntimeException are logged twice: once in `FBaseProcessor` and again by the Java uncaught exception handler.

@Workiva/messaging-pp

### DIFF
--- a/compiler/generator/java/generator.go
+++ b/compiler/generator/java/generator.go
@@ -3035,7 +3035,7 @@ func (g *Generator) generateServer(service *parser.Service) string {
 		contents += tabtabtabtab + "} catch (TException e) {\n"
 		contents += tabtabtabtabtab + "synchronized (WRITE_LOCK) {\n"
 		contents += tabtabtabtabtabtab + fmt.Sprintf(
-			"e = writeApplicationException(ctx, oprot, TApplicationExceptionType.INTERNAL_ERROR, \"%s\", \"Internal error processing %s: \" + e.getMessage());\n",
+			"e = (TApplicationException) writeApplicationException(ctx, oprot, TApplicationExceptionType.INTERNAL_ERROR, \"%s\", \"Internal error processing %s: \" + e.getMessage()).initCause(e);\n",
 			methodLower, method.Name)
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Raven
<details>

</details>

compiler/generator/java/generator.go
```diff
@@ -3035,7 +3035,7 @@ func (g *Generator) generateServer(service *parser.Service) string {
 		contents += tabtabtabtab + "} catch (TException e) {\n"
 		contents += tabtabtabtabtab + "synchronized (WRITE_LOCK) {\n"
 		contents += tabtabtabtabtabtab + fmt.Sprintf(
-			"e = writeApplicationException(ctx, oprot, TApplicationExceptionType.INTERNAL_ERROR, \"%s\", \"Internal error processing %s: \" + e.getMessage());\n",
+			"e = (TApplicationException) writeApplicationException(ctx, oprot, TApplicationExceptionType.INTERNAL_ERROR, \"%s\", \"Internal error processing %s: \" + e.getMessage()).initCause(e);\n",
```
brianshannan-wf: Now that I think about it more, do we want to expose these stack traces to clients? I think in the past we've erred on the side of trying to log as much as possible but restrict what we send back to the client somewhat

brettkail-wk: The stack trace doesn't get serialized back to the client, just the message.

brianshannan-wf: Ah, misread the parentheses there

brianshannan-wf:
+1

charliestrawn-wf:
lgtm

PR #955, opened 2018-01-08
Generator: update string for GAE python warning

aldenpeterson-wf:
@Workiva/messaging-pp 

This has caused a lot of confusion over the past few years. This updates the message to be more verbose.

@brettkail-wk @jeffthieleke-wf @matthewbelisle-wf

### DIFF
--- a/compiler/generator/python/generator.go
+++ b/compiler/generator/python/generator.go
@@ -997,7 +997,7 @@ func generatePrefixStringTemplate(scope *parser.Scope) string {
 // GenerateSubscriber generates the subscriber for the given scope.
 func (g *Generator) GenerateSubscriber(file *os.File, scope *parser.Scope) error {
 	// TODO
-	globals.PrintWarning(fmt.Sprintf("%s: scope subscriber generation is not implemented for Python", scope.Name))
+	globals.PrintWarning(fmt.Sprintf("%s: scope subscriber generation is not implemented for GAE Python, if you want this generate Tornado or Asyncio", scope.Name))
 	return nil
 }
 


aviary-wf:
## Raven
<details>

</details>

matthewbelisle-wf:
+1 thanks @aldenpeterson-wf

compiler/generator/python/generator.go
```diff
@@ -997,7 +997,7 @@ func generatePrefixStringTemplate(scope *parser.Scope) string {
 // GenerateSubscriber generates the subscriber for the given scope.
 func (g *Generator) GenerateSubscriber(file *os.File, scope *parser.Scope) error {
 	// TODO
-	globals.PrintWarning(fmt.Sprintf("%s: scope subscriber generation is not implemented for Python", scope.Name))
+	globals.PrintWarning(fmt.Sprintf("%s: scope subscriber generation is not implemented for GAE Python, if you want this generate Tornado or Asyncio", scope.Name))
```
tylertreat: IMO "GAE Python" isn't correct since this isn't a GAE-specific target. It's just plain Python without a concurrency model.

aldenpeterson-wf: I could change that wording, I put that in there given the use case here but I suppose in the generic sense that's true.

I'm not sure a good way to word it though.

jeffthieleke-wf: Maybe "...not implemented for vanilla Python 2.7.  For 2.7, use the Tornado framework (where available) or provide a pull request".

tylertreat: That sounds reasonable.

brianshannan-wf:
+1

charliestrawn-wf:
+1

PR #960, opened 2018-01-18
handle returning nil in a go middleware properly

brianshannan-wf:
If a client side middleware returns nil for the rpc return value (such as in the case of an error where the middleware doesn't necessarily know what type to return) we were trying to type assert nil to the proper type, causing a panic.

@Workiva/messaging-pp @Workiva/infre-product @tannermiller-wf


aviary-wf:
## Raven
<details>

</details>

stevenosborne-wf:
lgtm

PR #963, opened 2018-01-30
XS-6325 Bug fix for python code generation

sagesmith-wf:
# ULTIMATE PROBLEM:
Frugal generates python code that leads to sets and maps attempting to be hashed. This leads to failures when receiving an object that contains struct that has a set or a map as part of it.

# NOTE:
This is a simple fix until a better solution can be thought of that doesn't lose information when hashing lists that contain duplicate entries

# TESTING SUGGESTIONS:
Attempt to generate a struct that has a set and a map on it and make sure that calling the hash function on it doesn't break it. Example:

```thrift
struct PythonCollectionTesterDTO {
    1: string aNonSpecialCase
    2: set<string> aSet
    3: list<string> aList
    4: map<string, string> aMap
}
```

Try something along these lines and check if there is an error:

```python

from myStruct import PythonCollectionTesterDTO

if __name__ == "__main__" :
    test_obj = PythonCollectionTesterDTO()
    test_obj.aNonSpecialCase = "not special"
    test_obj.aSet = set()
    test_obj.aList = []
    test_obj.aMap = dict()
    test_obj.__hash__()
```
# AREAS AFFECTED:
Compiler

Please Review: @Workiva/messaging-pp

### DIFF
--- a/compiler/generator/python/generator.go
+++ b/compiler/generator/python/generator.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"frozendict"
 
 	"github.com/Workiva/frugal/compiler/generator"
 	"github.com/Workiva/frugal/compiler/globals"
@@ -499,7 +500,13 @@ func (g *Generator) generateMagicMethods(s *parser.Struct) string {
 	contents += tab + "def __hash__(self):\n"
 	contents += tabtab + "value = 17\n"
 	for _, field := range s.Fields {
-		contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(self.%s)\n", field.Name)
+	    if field.Type.KeyType != nil { // Check if map, hash frozen map in that case
+	        contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(frozendict(self.%s))\n", field.Name)
+	    } else if field.Type.ValueType != nil { // Check if set or list, hash frozenset in that case. Note: Hash may be incorrect for lists with duplicates
+	        contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(frozenset(self.%s))\n", field.Name)
+	    } else { // Write normally
+		    contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(self.%s)\n", field.Name)
+		}
 	}
 	contents += tabtab + "return value\n\n"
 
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Raven
<details>

</details>

## Magpie
<details>
<summary>Number of findings: 0</summary>

</details>

scripts/smithy/smithy_python.sh
```diff
@@ -16,6 +16,7 @@ make xunit-py2
 pip freeze > $SMITHY_ROOT/python2_pip_deps.txt
 $FRUGAL_HOME/scripts/smithy/codecov.sh $FRUGAL_HOME/lib/python/unit_tests_py2.xml python_two
 make flake8-py2
+make frozendict-py2
```
aldenpeterson-wf: is this missing a makefile include as well?

compiler/generator/python/generator.go
```diff
@@ -499,7 +499,14 @@ func (g *Generator) generateMagicMethods(s *parser.Struct) string {
 	contents += tab + "def __hash__(self):\n"
 	contents += tabtab + "value = 17\n"
 	for _, field := range s.Fields {
-		contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(self.%s)\n", field.Name)
+	    if field.Type.KeyType != nil { // Check if map, hash frozen map in that case
```
aldenpeterson-wf: nit: this is easier to read if the comment is a separate line instead of inline

theisensanders-wf:
```
struct PythonCollectionTesterDTO {
    1: string aNonSpecialCase
    2: set<string> aSet
    3: list<string> aList
    4: map<string, list> aMap
    5: binary someBytes
}
```

Would like to make sure it behaves with a list as a map value (or some other way of causing recursive hashing) and I think binary is a `bytearray` in python which is also not hashable (but can be converted to `bytes`)

compiler/generator/python/generator.go
```diff
@@ -499,7 +499,14 @@ func (g *Generator) generateMagicMethods(s *parser.Struct) string {
...
+	    	contents += "import frozendict\n"
+	        contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(frozendict(self.%s))\n", field.Name)
+	    } else if field.Type.ValueType != nil { // Check if set or list, hash frozenset in that case. Note: Hash may be incorrect for lists with duplicates
+	        contents += fmt.Sprintf(tabtab+"value = (value * 31) ^ hash(frozenset(self.%s))\n", field.Name)
```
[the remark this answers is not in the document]
sagesmith-wf: I don't really see a good way of differentiating between a list and a set at this time, does anyone have any ideas?

PR #969, opened 2018-02-19
java: Add FContext methods to remove headers

brettkail-wk:
It is useful to be able to clone an FContext in order to clone correlation id, tracing, etc. headers but to remove others.

@Workiva/messaging-pp

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/FContext.java
+++ b/lib/java/src/main/java/com/workiva/frugal/FContext.java
@@ -159,6 +159,19 @@ public class FContext implements Cloneable {
     }
 
     /**
+     * Removes a request header from the FContext for the given name.
+     * The _opid header is reserved and removing it could interfere with asynchronous transports that rely upon it.
+     * Returns the same FContext to allow for call chaining.
+     *
+     * @param name header name
+     * @return FContext
+     */
+    public FContext removeRequestHeader(String name) {
+        requestHeaders.remove(name);
+        return this;
+    }
+
+    /**
      * Adds a response header to the FContext for the given name. A header is a key-value pair.
      * If a header with the name is already present on the FContext, it will be replaced.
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
<details>

</details>

## Magpie
<details>
<summary>Number of findings: 0</summary>

</details>

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -158,6 +158,19 @@ public FContext addRequestHeaders(Map<String, String> headers) {
         return this;
     }
 
+    /**
+     * Removes a request header from the FContext for the given name.
+     * The _opid header is reserved and removing it could interfere with asynchronous transports that rely upon it.
```
aldenpeterson-wf: Is it worth explicitly checking for this?

I'm never sure how much to code against fairly big mistakes vs just... assuming people do things right.

aldenpeterson-wf: Same with correlation ids, right?

aldenpeterson-wf:
lgtm, I'll let @brianshannan-wf chime in whether it's worth explicitly erroring in cases of removing any reserved headers

lib/java/src/main/java/com/workiva/frugal/FContext.java
brettkail-wk: This is copy/paste from the add methods.  I'm happy to leave, remove, or remove from all, just let me know what you all prefer.

brianshannan-wf: I'm fine not explicitly checking for it, it follows the convention we've laid out in other methods

brianshannan-wf:
+1

stevenosborne-wf:
+1

PR #971, opened 2018-03-01
Http gateway prototype

kevinsookocheff-wf:
@brianshannan-wf @shawnrusaw-wf 

Initial prototype for an HTTP+JSON proxy.  The README and exampleGateway.go in this PR should show how you would use it.

### New Code

- A Frugal compiler addition reads IDL with http annotations and creates an HTTP handler for each RPC method.
- A Frugal library addition that handles common serialization tasks and HTTP server runtime behaviour

### Extended Code

- The gateway compiler generator has a lot of common functionality with the go compiler generator so I made some methods in the go compiler public for re-use.


aviary-wf:
## Module Findings

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+
+	contents += g.GenerateInlineComment([]string{fmt.Sprintf("%s forwards HTTP requests to a Frugal service", contextName)}, "")
+	contents += fmt.Sprintf("type %s struct {\n", contextName)
+	contents += fmt.Sprintf("\tClient *F%sClient\n", serviceTitle)
+	contents += "\tMarshalers gateway.MarshalerRegistry\n"
+	contents += "}\n\n"
+
+	contents += g.GenerateInlineComment([]string{fmt.Sprintf("%sHandler is a wrapper to provide context to HTTP handlers", serviceTitle)}, "")
+	contents += fmt.Sprintf("type %sHandler struct {\n", serviceTitle)
+	contents += fmt.Sprintf("\t*%s\n\n", contextName)
```
brianshannan-wf: Is there a reason you embedded this versus having it as a field? To me embedding is an "is" relationship like inheritance and the handler doesn't seem like it "is" a context

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+	contents += "\t}\n"
+	contents += "\n}\n\n"
+
+	return contents
+}
+
+func (g *Generator) generateFieldSetter(conversion string, fieldName string) string {
+	var contents = ""
+
+	contents += "\t\t\t\t\tc, err := gateway.String(v)\n"
```
brianshannan-wf: I think this line should be replaced with `conversion`?

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+	contents += "\t\t\t\t\tif err != nil {\n"
+	contents += "\t\t\t\t\t\tpanic(err)\n"
```
brianshannan-wf: I think this error can be returned as the enclosing function returns an error

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+	contents += "\t\t\t\t\t}\n"
+
+	contents += fmt.Sprintf("\t\t\t\t\tf := s.Field(\"%s\")\n", fieldName)
+	contents += "\t\t\t\t\tif strings.Contains(f.Tag(\"json\"), \"omitempty\") {\n"
+	contents += "\t\t\t\t\t\terr = f.Set(&c)\n"
+	contents += "\t\t\t\t\t} else {\n"
+	contents += "\t\t\t\t\t\terr = f.Set(c)\n"
```
brianshannan-wf: This error should probably be handled somehow

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+
+func (g *Generator) generateFieldSetter(conversion string, fieldName string) string {
+	var contents = ""
+
+	contents += "\t\t\t\t\tc, err := gateway.String(v)\n"
+	contents += "\t\t\t\t\tif err != nil {\n"
+	contents += "\t\t\t\t\t\tpanic(err)\n"
+	contents += "\t\t\t\t\t}\n"
+
+	contents += fmt.Sprintf("\t\t\t\t\tf := s.Field(\"%s\")\n", fieldName)
```
brianshannan-wf: I guess I'm not sure why you're creating these fields? It seems like you could set attributes directly on generated struct you create, `payload`

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+	// If no HTTP annotation, return without generating handler
+	_, found := method.Annotations.Get("http.pathTemplate")
+	if !found {
+		return ""
+	}
+
+	contents += g.GenerateInlineComment([]string{fmt.Sprintf("%s%sHandler forwards HTTP requests to a Frugal service", serviceTitle, methodTitle)}, "")
+	contents += fmt.Sprintf("func %s%sHandler(context *%s, responseWriter http.ResponseWriter, request *http.Request) (int, error) {\n", serviceTitle, methodTitle, contextName)
+
+	contents += "\tflusher, _ := responseWriter.(http.Flusher)\n\n"
```
brianshannan-wf: Should either handle the returned error or use the no error panic version of the cast

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+	contents += "\terr := decoder.Decode(payload)\n"
+	contents += "\tif err != nil && err != io.EOF {\n"
+	contents += "\t\tpanic(err) // TODO: Customize error handling\n"
+	contents += "\t}\n\n"
+
+	contents += g.GenerateInlineComment([]string{"Combine path and query parameters into map[string]string.", "If there are duplicate query parameters, only the first is respected."}, "\t")
+	contents += "\tvars := mux.Vars(request)\n"
+	contents += "\tqueries := request.URL.Query()\n"
+	contents += "\tfor k, v := range queries {\n"
+	contents += "\t	vars[k] = v[0]\n"
```
brianshannan-wf: should be `\t\t` instead of `\t `

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+	parsedStruct := g.Frugal.FindStruct(argument.Type)
+	contents += "\ts := gateway.NewStruct(payload)\n"
+	contents += "\t\tfor k, v := range vars {\n"
+	for _, field := range parsedStruct.Fields {
+		fieldName := field.Name
+		if jsonProperty, ok := field.Annotations.Get("http.jsonProperty"); ok {
+			fieldName = jsonProperty
+		}
+		contents += fmt.Sprintf("\t\t\t\tif k == \"%s\" {\n", fieldName)
+		switch field.Type.String() {
```
brianshannan-wf: If we already know the field name, can we just typecast to the type we expect it to be?

compiler/generator/gateway/generator.go
```diff
@@ -0,0 +1,403 @@
...
+		}
+		contents += "\t\t\t\t}\n"
+	}
+	contents += "\t\t}\n\n"
+
+	// Pass the payload to the Frugal client
+	contents += g.GenerateInlineComment([]string{"Call the Frugal client with the assembled payload"}, "\t")
+	contents += fmt.Sprintf("\tresponse, err := context.Client.%s(frugal.NewFContext(\"\"), payload)\n", methodTitle)
+	contents += "\tif err != nil {\n"
+	contents += "\t\tpanic(err) // TODO: Customize error handling\n"
```
brianshannan-wf: Is there any error handling that needs to be done here other than returning the error for the caller to deal with?

lib/gateway/errors.go
```diff
@@ -0,0 +1,174 @@
...
+// NewValidationError creates a new Validation error for reporting back to the user
+func NewValidationError(resource string, field string, code string) *ValidationError {
+	return &ValidationError{resource, field, code}
+}
+
+// HTTPStatusFromFrugalError converts a Frugal exception into a corresponding HTTP response status.
+// https://github.com/Workiva/frugal/blob/master/lib/go/errors.go
+func HTTPStatusFromFrugalError(err thrift.TException) int {
+	fmt.Println("http status from frugal error")
+	switch err.(type) {
```
brianshannan-wf: you can do `switch e := err.(type)` to avoid recasting multiple times

lib/gateway/errors.go
```diff
@@ -0,0 +1,174 @@
...
+	// Write the response
+	w.Header().Set("Content-Type", marshaler.ContentType())
+	w.WriteHeader(status)
+	w.Write(buf)
+	f.Flush()
+}
+
+// DefaultResponseErrorHandler is the default implementation of an error handler for outgoing HTTP responses.
+func DefaultResponseErrorHandler(marshaler Marshaler, w http.ResponseWriter, r *http.Request, err error) {
+
```
brianshannan-wf: I assume this is one of the TODO's?

lib/gateway/exampleGateway/main.go
```diff
@@ -0,0 +1,68 @@
+package main
```
brianshannan-wf: Should this go in `examples/gateway` or something similar?

lib/gateway/gateway_test.go
```diff
@@ -0,0 +1,79 @@
+// These tests require a running server and gateway before running.
```
brianshannan-wf: It seems like these belong in the skynet tests if they're more integration style

lib/gateway/marshal_json_test.go
```diff
@@ -0,0 +1,195 @@
...
+var strPtr = &str
+
+func TestJSONBuiltinMarshalUnmarshal(t *testing.T) {
+	// Given
+	var m JSONBuiltin
+	msg := marshal_test.BaseType{StringTest: strPtr}
+
+	// When
+	buf, err := m.Marshal(&msg)
+	if err != nil {
```
brianshannan-wf: You can use functions from `github.com/stretchr/testify/assert` for doing this easier, such as `assert.NoError(t, err)`

compiler/generator/gateway/generator.go
kevinsookocheff-wf: Error handling needs work all around. Consider error handling incomplete right now.

compiler/generator/gateway/generator.go
kevinsookocheff-wf: There might be a better way to do this (probably there is).

I had trouble setting the fields directly because I had to convert from string to a certain value at runtime without knowing what the correct type will be until runtime.

lib/gateway/errors.go
kevinsookocheff-wf: Yes. What I would like to do is have an "error JSON template" that is configurable by the user and then the error handler would fill in the JSON template with what it reports back.

PR #972, opened 2018-03-05
Python: import subpackage objects in __init__.py

aaronkavlie-wf:
Description
========
It's common in Python code to pull imports into an `__init__.py` file — see [here](https://github.com/pallets/flask/blob/master/flask/__init__.py#L21-L31) and [here](https://github.com/requests/requests/blob/master/requests/__init__.py#L112-L122), for example. This makes the public API easily accessible with short, simple import paths.

In that vein, this PR adds imports to `__init__.py` files in the Python codegen.

**Before:**

```python
from tasker_frugal_v2.asyncio.f_TaskIntegrationService import Client as TaskIntegrationClient
from workflow_frugal.asyncio.v1.workflow_events.f_FWorkflowEgress_subscriber \
    import FWorkflowEgressSubscriber
from workflow_frugal.asyncio.v1.workflow_events.f_FWorkflowIngress_publisher \
    import FWorkflowIngressPublisher
from workflow_frugal.asyncio.v1.workflow_events.ttypes \
    import FWorkflowUserTaskEvent
```

**After:**

```python
from tasker_frugal_v2.asyncio import TaskIntegrationServiceClient
from workflow_frugal.asyncio.v1.workflow_events import (
    FWorkflowEgressSubscriber, FWorkflowIngressPublisher, FWorkflowUserTaskEvent)
```

Code Review
=========
@brianshannan-wf @charliestrawn-wf


aviary-wf:
## Module Findings

aaronkavlie-wf:
All checks passed

compiler/generator/python/generator.go
```diff
@@ -68,6 +71,7 @@ func (g *Generator) SetupGenerator(outputDir string) error {
 
 	dir := g.outputDir
 	for filepath.Dir(dir) != "." {
+		fmt.Println(dir)
```
brianshannan-wf: This should be removed

compiler/generator/python/generator.go
```diff
@@ -101,9 +105,70 @@ func (g *Generator) SetupGenerator(outputDir string) error {
...
+		return err
+	}
+
 	return g.typesFile.Close()
 }
 
+// generateInit generates __init__.py files, with subpackage imports
+// to simplify consumer import paths
+func (g *Generator) generateInitFile() error {
+	initFile, err := os.Create(path.Join(g.outputDir, "__init__.py"))
```
brianshannan-wf: Isn't this already created?

compiler/generator/python/generator.go
```diff
@@ -101,9 +105,70 @@ func (g *Generator) SetupGenerator(outputDir string) error {
...
+	defer initFile.Close()
+
+	initWriter := bufio.NewWriter(initFile)
+
+	dir, err := os.Open(g.outputDir)
+	if err != nil {
+		return err
+	}
+
+	names, err := dir.Readdirnames(0)
```
brianshannan-wf: While I think this would work for almost every case, I'm not a huge fan of this. For example, if there's other things in this directory it may not work as intended. Is it possible to either loop through `g.Frugal.Structs` etc. to get this or accumulate the file contents as a string during the generation process? It seems like that would be a more robust solution

compiler/generator/python/generator.go
aaronkavlie-wf: So at first I removed the earlier loop that created `__init__.py` files, but found that that loop covered some critical files that aren't created here.

Go's APIs for writing a pre-existing file are awkward; I can switch to an [os.OpenFile()](https://golang.org/pkg/os/#OpenFile) call. It shouldn't make a difference in end result, but would clarify the code a bit.

compiler/generator/python/generator.go
aaronkavlie-wf: I'll take a look at `g.Frugal.Structs` to see if that's a viable solution.

I thought I'd be able to do this by looking at some internal bookkeeping, but I couldn't find any avenues for that without some larger code changes that would have required changes beyond the Python generator. I agree that inferring from file names is not ideal.

aaronkavlie-wf:
@brianshannan-wf Updated with an approach based on your suggestion; it avoids reading file names now.

I think this is simpler and cleaner than my first approach.

compiler/generator/python/generator.go
```diff
@@ -101,9 +110,51 @@ func (g *Generator) SetupGenerator(outputDir string) error {
...
+				imports = append(imports, fmt.Sprintf("from .%s import %sPublisher", fileInfo.fileName, fileInfo.frugalName))
+			case generator.SubscribeFile:
+				imports = append(imports, fmt.Sprintf("from .%s import %sSubscriber", fileInfo.fileName, fileInfo.frugalName))
+			case generator.CombinedServiceFile:
+				imports = append(imports,
+					fmt.Sprintf("from .%s import Iface as F%sIface", fileInfo.fileName, fileInfo.frugalName))
+				imports = append(imports,
+					fmt.Sprintf("from .%s import Client as F%sClient", fileInfo.fileName, fileInfo.frugalName))
+			case generator.ObjectFile:
+				if fileInfo.frugalName == "ttypes" {
```
brianshannan-wf: IIRC this won't import constants as ttypes doesn't import constants. We might want to do constants instead as I think it imports ttypes so we'd get ttypes transitively

aaronkavlie-wf: That's the first approach I tried actually, but it led to circular imports.
I can try again and paste the stack trace if it helps.

brianshannan-wf: Oh yeah, forgot about that

brianshannan-wf:
+1

brianshannan-wf:
@Workiva/messaging-pp @Workiva/infre-product

garrickpeterson-wf:
+1

The separation of the import file opening and the file writing doesn't read that cleanly; but it's an inconsequential nit.

PR #973, opened 2018-03-06
java: Avoid comment for default requiredness

brettkail-wk:
It is confusing to output a `// required` comment for fields that have default requiredness.

@Workiva/messaging-pp


aviary-wf:
## Module Findings

brianshannan-wf:
+1. I think those comments were a remnant of duplicating code thrift produced. This does change the presence modifier passed to the type function, but I don't think that should change anything

charliestrawn-wf:
+1

stevenosborne-wf:
+1

PR #977, opened 2018-03-12
Fix vendor folder. Use glide, but keep godep around for people using godep commands

charliestrawn-wf:


aviary-wf:
## Module Findings

brianshannan-wf:
@Workiva/messaging-pp @Workiva/infre-product

charliestrawn:
+1

garrickpeterson-wf:
This is a huge diff, but it looks like we are still checking in the dependencies? I thought the entire premise of using glide was not having to check them in to get consistent builds.

Speaking only from an infre point of view - when we have glide dependencies, we do not check the vendor directory into git.

brianshannan-wf:
Note: please don't review until I'm able to test all the godep commands still work properly

brianshannan-wf:
We need to keep the vendor folder for now due to godep. We can remove this once the deprecation period is over, which we need to decide on and communicate out

garrickpeterson-wf:
+1

tylerrinnan-wf:
+1

PR #979, opened 2018-03-15
Fix quotes in annotation string literals

brettkail-wk:
@Workiva/messaging-pp

### DIFF
--- a/compiler/generator/golang/generator.go
+++ b/compiler/generator/golang/generator.go
@@ -203,6 +203,11 @@ func (g *Generator) GenerateConstantsContents(constants []*parser.Constant) erro
 	return nil
 }
 
+// quote creates a Go string literal for a string.
+func (g *Generator) quote(s string) string {
+	return strconv.Quote(s);
+}
+
 // generateConstantValue recursively generates the string representation of
 // a, possibly complex, constant value.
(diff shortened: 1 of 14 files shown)


aviary-wf:
## Module Findings

compiler/generator/golang/generator.go
```diff
@@ -1856,7 +1861,7 @@ func (g *Generator) generateProcessor(service *parser.Service) string {
 		if len(method.Annotations) > 0 {
 			contents += fmt.Sprintf("\tp.AddToAnnotationsMap(\"%s\", map[string]string{\n", methodLower)
 			for _, annotation := range method.Annotations {
-				contents += fmt.Sprintf("\t\t\"%s\": \"%s\",\n", annotation.Name, annotation.Value)
+				contents += fmt.Sprintf("\t\t%s: %s,\n", g.quote(annotation.Name), g.quote(annotation.Value))
```
brianshannan-wf: Do we want annotation names to be able to contain quotes? I guess that seems kind of weird to me but I could be convinced

brettkail-wk: You're right, annotation keys must be identifiers, so we don't need to quote them.

brianshannan-wf:
+1

charliestrawn-wf:
lgtm

PR #982, opened 2018-03-21
java: Add FServlet

brettkail-wk:
The existing `FNettyHttpHandler` works well, but the asynchronous nature of the Netty programming model is more difficult to use than the servlet programming model for non-Frugal use cases.  Add a servlet so that Frugal can easily be included in the same server as other servlets.

@Workiva/messaging-pp

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
@@ -24,6 +24,9 @@ import java.util.Base64;
 /**
  * Processes POST requests as Frugal requests for a processor.
  * <p>
+ * By default, the HTTP request is limited to a 64MB Frugal payload size to
+ * prevent client requests from causing the server to allocate too much memory.
+ * <p>
  * The HTTP request may include an X-Frugal-Payload-Limit header setting the size
  * limit of responses from the server.
  * <p>
@@ -37,7 +40,7 @@ import java.util.Base64;
 public class FServlet extends HttpServlet {
     private static final Logger LOGGER = LoggerFactory.getLogger(FServlet.class);
 
-    private static final int DEFAULT_MAX_REQUEST_SIZE = 1024 * 1024;
+    private static final int DEFAULT_MAX_REQUEST_SIZE = 64 * 1024 * 1024;
 
     private final FProcessor processor;
     private final FProtocolFactory inProtocolFactory;
--- a/lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
@@ -1,0 +1,160 @@
+package com.workiva.frugal.server;
+
+import com.workiva.frugal.processor.FProcessor;
+import com.workiva.frugal.protocol.FProtocolFactory;
+import com.workiva.frugal.transport.TMemoryOutputBuffer;
+import org.apache.thrift.TException;
+import org.apache.thrift.transport.TMemoryInputTransport;
+import org.apache.thrift.transport.TTransport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import java.io.DataInputStream;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Base64;
+
+/**
+ * Processes POST requests as Frugal requests for a processor.
+ * <p>
+ * The HTTP request may include an X-Frugal-Payload-Limit header setting the size
+ * limit of responses from the server.
+ * <p>
+ * The HTTP processor returns a 500 response for any runtime errors when executing
+ * a frame, a 400 response for an invalid frame, and a 413 response if the response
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Module Findings

lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
```diff
@@ -0,0 +1,160 @@
...
+ * a frame, a 400 response for an invalid frame, and a 413 response if the response
+ * exceeds the payload limit specified by the client.
+ * <p>
+ * Both the request and response are base64 encoded.
+ */
+@SuppressWarnings("serial")
+public class FServlet extends HttpServlet {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FServlet.class);
+
+    private static final int DEFAULT_MAX_REQUEST_SIZE = 1024 * 1024;
```
brianshannan-wf: We don't impose a max request size by default in other languages (and is one of the reasons the http transport exists), so this may cause some confusion when compared to other languages http server implementation. That said, I'm not opposed to sane defaults for message sizes. This should probably be documented if we stick with it though

brianshannan-wf:
In addition to unit tests we have some integration tests for transports/servers that runs in skynet, https://github.com/Workiva/frugal/tree/master/test/integration. It might be worth adding this to that. I can probably do that if needed.

lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
brettkail-wk: I had copied from the Netty code, and I thought the `HttpObjectAggregator` limit was builtin, but I guess it's only in example code.  I would still prefer to some limit rather than potentially allocating a 2GB array.  Is 64MB a reasonable default?  Alternatively, I could use `TIOStreamTransport` for the input, but the error handling would be a bit more complicated, and it didn't look like other languages were doing that.

kevinsookocheff-wf:
@brettkail-wk Did somebody ask for a servlet implementation? I had done one before but that PR was closed because nobody was running servlet containers within docker containers.

brettkail-wk:
@kevinsookocheff-wf We have services that would like to use servlet + Frugal.  If you already have code that you would prefer to use instead, then go ahead and merge that and close this.

kevinsookocheff-wf:
The code I wrote is probably > year old now so let's go with this. I just wanted to make sure people were running a servlet-based server.

lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
brianshannan-wf: I'm fine with limits, as long as it's documented and configurable

brettkail-wk:
Opened #983 for adding the integration test.

brianshannan-wf:
+1

lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public FServlet(
+            FProcessor processor,
+            FProtocolFactory inProtocolFactory,
+            FProtocolFactory outProtocolFactory,
+            int maxRequestSize) {
+        this.processor = processor;
+        this.inProtocolFactory = inProtocolFactory;
+        this.outProtocolFactory = outProtocolFactory;
+        this.maxRequestSize = maxRequestSize;
```
garrickpeterson-wf: Is 64MB a hard enough limit that we should we be adding a check against `DEFAULT_MAX_REQUEST_SIZE` here (so someone can instantiate a class with a lower limit, but not a higher)?

brettkail-wk: No.  It's an arbitrary number bigger than the default of NATS (since that's one common reason to use HTTP) but small enough to probably avoid OOM for common server configurations.  A server needs to ensure that maxThreads * maxRequestSize < maxHeap + spaceForWorking, so a server that configures a low number of threads can safely choose a higher maxRequestSize or vice versa.

lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
```diff
@@ -0,0 +1,163 @@
...
+        int responseLimit = getResponseLimit(req);
+        if (responseLimit > 0 && outTransport.size() > responseLimit) {
+            LOGGER.debug("Response size too large for client. Received: {}, Limit: {}",
+                    outTransport.size(), responseLimit);
+            resp.setStatus(HttpServletResponse.SC_REQUEST_ENTITY_TOO_LARGE);
+            return;
+        }
+
+        resp.setContentType("application/x-frugal");
+        resp.setHeader("Content-Transfer-Encoding", "base64");
```
garrickpeterson-wf: Do we want to set the `Content-Length` header as well, since we know the size?

tylerrinnan-wf:
+1

lib/java/src/main/java/com/workiva/frugal/server/FServlet.java
brettkail-wk: I considered it, but we would have to compute the base64-encoded size, and I didn't think it was worth the effort (or have yet another base64-encoded copy of the data in memory, which seems worse).

brettkail-wk: I should say: Good question, and I should have included this detail in my PR since I actually had the code written to compute the base64-encoded size and set Content-Length but didn't think it was worth the complexity :-).

PR #985, opened 2018-03-26
Support deprecated annotation for fields

brettkail-wk:
The support varies depending on the language:
- Dart: Add the `@deprecated` annotation, but only add "/// Deprecated" doc comments to the get/set for fields to avoid too much reptition.
- Go: Append "Deprecated" to the comments of the exported field. Additionally, add the deprecation value to service method comments.
- HTML: Append "Deprecated" to the comments column of the field.
- Java: Add the `@Deprecated` annotation to the public fields and all field-related methods, but only add javadoc to fields and get/set methods to avoid too much reptition.
- Python: Append "Deprecated" to the comment for the field in the class docstring.  Additionally, indent second and later comment lines so that it does not appear the following field is deprecated.

@Workiva/messaging-pp

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Some common annotations are listed below
 | Annotation    | Values        | Allowed Places | Description
 | ------------- | ------------- | -------------- | -----------
 | vendor        | Optional location | Namespaces, Includes | See [vendoring includes](#vendoring-includes)
-| deprecated    | Optional description | Service methods | Marks a method as deprecated (if supported by the language) and logs a warning if the method is called.
+| deprecated    | Optional description | Service methods, Struct/union fields | Marks a method or field as deprecated (if supported by the language, or in a comment otherwise), and logs a warning if a deprecated method is called.
 
 ### Vendoring Includes
 
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Some common annotations are listed below
 | Annotation    | Values        | Allowed Places | Description
 | ------------- | ------------- | -------------- | -----------
 | vendor        | Optional location | Namespaces, Includes | See [vendoring includes](#vendoring-includes)
-| deprecated    | Optional description | Service methods, Struct/union fields | Marks a method or field as deprecated (if supported by the language, or in a comment otherwise), and logs a warning if a deprecated method is called.
+| deprecated    | Optional description | Service methods, Struct/union/exception fields | Marks a method or field as deprecated (if supported by the language, or in a comment otherwise), and logs a warning if a deprecated method is called.
 
 ### Vendoring Includes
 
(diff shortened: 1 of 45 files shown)


aviary3-wk:
## Module Findings

README.md
```diff
@@ -215,7 +215,7 @@ Some common annotations are listed below
 | Annotation    | Values        | Allowed Places | Description
 | ------------- | ------------- | -------------- | -----------
 | vendor        | Optional location | Namespaces, Includes | See [vendoring includes](#vendoring-includes)
-| deprecated    | Optional description | Service methods | Marks a method as deprecated (if supported by the language) and logs a warning if the method is called.
+| deprecated    | Optional description | Service methods, Struct/union fields | Marks a method or field as deprecated (if supported by the language, or in a comment otherwise), and logs a warning if a deprecated method is called.
```
brianshannan-wf: Should this work for exception fields as well? They're very similar to structs/unions

brianshannan-wf:
+1

compiler/generator/dartlang/generator.go
```diff
@@ -1511,6 +1508,22 @@ func (g *Generator) GenerateService(file *os.File, s *parser.Service) error {
 	return err
 }
 
+func (g *Generator) generateCommentWithDeprecated(comment []string, indent string, anns parser.Annotations) string {
+	contents := ""
+	if comment != nil {
+		contents += g.GenerateInlineComment(comment, tab+"/")
```
garrickpeterson-wf: These don't seem to be using the indent, just a constant `tab`. Is this the standard for dart?

brianshannan-wf: It would still produce the correct output right now because only `tab` is passed in but this should be changed to use `indent` so if we add something like class deprecation in the future that'll work properly. Good catch

compiler/generator/dartlang/generator.go
```diff
@@ -1511,6 +1508,22 @@ func (g *Generator) GenerateService(file *os.File, s *parser.Service) error {
...
+	}
+
+	if deprecationValue, deprecated := anns.Deprecated(); deprecated {
+		if deprecationValue != "" {
+			contents += g.GenerateInlineComment([]string{"Deprecated: " + deprecationValue}, tab+"/")
```
garrickpeterson-wf: It would be nice if there was a different method to generate dart doc comments instead of just the `tab+"/"` as an indent. Probably a change that's out of scope of this fix.

brettkail-wk: Indeed, this one I'm going to leave alone for now :-).

compiler/generator/golang/generator.go
```diff
@@ -442,6 +442,24 @@ func (g *Generator) generateStruct(s *parser.Struct, serviceName string) string
...
+	if comment != nil {
+		contents += g.GenerateInlineComment(comment, indent)
+	}
+
+	deprecationValue, deprecated := anns.Deprecated()
+	if deprecated && deprecationValue != "" {
+		if deprecationValue == "" {
+			contents += indent + "// Deprecated\n"
+		} else {
+			contents += fmt.Sprintf("%s// Deprecated: %s\n", indent, deprecationValue)
```
garrickpeterson-wf: Should this be a doc comment?

garrickpeterson-wf:
+1

brianshannan-wf:
+1

PR #988, opened 2018-04-06
Fix travis for forks and dont try to push artifacts if not Workiva repo

charliestrawn:
Forks do some funky stuff with the gopath since it clones into user/repo. Just moving the repo to the upstreams checkout folder appears to fix it. 

Also the script was trying to push on commits to master of my fork. This obviously fails because my travis fork doesn't have the necessary encrypted vars, but we should still just skip trying to deploy if not Workiva/frugal.

@brianshannan-wf @tylerrinnan-wf

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ env:
   - TWINE_REPOSITORY_URL=https://upload.pypi.org/latest/
 
 before_install:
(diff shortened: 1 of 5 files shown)


aviary-wf:
## Module Findings

brianshannan-wf:
+1

brianshannan-wf:
It looks like codecov needs fixed

PR #992, opened 2018-04-14
Fix travis builds for forks, don't publish if not Workiva/frugal, and add a top level makefile.

charliestrawn:
@brianshannan-wf @tylerrinnan-wf


aviary-wf:
## Module Findings

brianshannan-wf:
+1

PR #993, opened 2018-04-18
Glide examples

brianshannan-wf:
- Added vendoring to the go example so it was easier to run
- Added a makefile that would sym link the local frugal dependency so they could be easily tested in the examples
- Added group mentions to the pr template

### DIFF
--- a/examples/go/Makefile
+++ b/examples/go/Makefile
@@ -1,0 +1,13 @@
+
+.PHONY: all
+all: install
+
+# Installs dependencies while sym linking the local frugal into the
+# vendor folder so local changes can be easily tested
+.PHONY: install
+install:
+	rm -rf vendor && \
+		glide install && \
+		rm -rf vendor/github.com/Workiva/frugal/lib/go && \
+		ln -s ${GOPATH}/src/github.com/Workiva/frugal/lib/go vendor/github.com/Workiva/frugal/lib/go
+
--- a/examples/go/glide.yaml
+++ b/examples/go/glide.yaml
@@ -1,0 +1,22 @@
+package: github.com/Workiva/frugal/examples/go
+import:
+- package: git.apache.org/thrift.git
+  subpackages:
+  - lib/go/thrift
+  version: 0.10.0
+- package: github.com/Sirupsen/logrus
+  repo: git@github.com:/sirupsen/logrus
+  vcs: git
+  version: ^v1.0.5
+- package: github.com/Workiva/frugal
+  subpackages:
+  - lib/go
+  version: 2.17.0
+- package: github.com/nats-io/go-nats
+  version: 6b6bf392d34d01f57cc563ae123f00c13778bd57
+- package: github.com/rs/cors
+  version: v1.3.0
+- package: golang.org/x/sys
+  version: b126b21c05a91c856b027c16779c12e3bf236954
+  subpackages:
+  - unix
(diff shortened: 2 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

examples/go/Makefile
```diff
@@ -0,0 +1,13 @@
+
+.PHONY: all
+all: install
+
+# Installs dependencies while sym linking the local frugal into the
+# vendor folder so local changes can be easily tested
```
aldenpeterson-wf: nit: should this say "local changes to frugal" ? I think that's what you mean but it reads a bit awkard.

examples/go/Makefile
```diff
@@ -0,0 +1,13 @@
...
+.PHONY: install
+install:
+	rm -rf vendor && \
+		glide install && \
+		rm -rf vendor/github.com/Workiva/frugal/lib/go && \
+		ln -s ${GOPATH}/src/github.com/Workiva/frugal/lib/go vendor/github.com/Workiva/frugal/lib/go
```
aldenpeterson-wf: This could be fun if someone doesn't have GOPATH configured 😈

examples/go/glide.yaml
```diff
@@ -0,0 +1,22 @@
+package: github.com/Workiva/frugal/examples/go
```
aldenpeterson-wf: does the update script need to update this, too?

brianshannan-wf: Why would that need updated? Or what would it be updated to?

examples/go/Makefile
brianshannan-wf: I meant this makefile for people to use for development, in which case they should have their gopath configured. In fact, I don't think vendor folders work outside of a gopath.

charliestrawn-wf: I don't think you need the { }

brianshannan-wf: I definitely did, otherwise make thought `$G` was the variable and symlinked `OPATH/src/github.com/Workiva/frugal/lib/go`.

charliestrawn-wf:
lgtm

examples/go/glide.yaml
aldenpeterson-wf: Oh. I guess I had assumed there would be overlap here with the overall glide.yaml https://github.com/Workiva/frugal/blob/master/glide.yaml but I guess not.

The version on line 14 seems that way, though as it explicitly is pinned to the version of frugal and that seems maybe it should be in sync - though I'm not sure of that.

brianshannan-wf: Oh, I see what you mean. Yeah I can do that

briansparrow-wk:
+1

scripts/lang/go.py
```diff
@@ -11,7 +14,16 @@ def update_frugal(self, version, root):
         Update the go version. Go versioning is controlled by git tags, so
         there is nothing to do here.
         """
-        pass
+        os.chdir('{0}/examples/go'.format(root))
+
+        with open('glide.yaml') as f:
+            glide_yaml = yaml.load(f.read())
```
kevinfrommelt-wf: Can you change `load()` to `safe_load()`

kevinfrommelt-wf:
+10

charliestrawn-wf:
lgtm

PR #1000, opened 2018-05-10
Go gen new args result

brianshannan-wf:
https://jira.atl.workiva.net/browse/INFENG-5682

Top level names are conditionally formatted with a `_` at the end of the name if it would start with `New` or end with `Args` or `Result`. This is to avoid conflicts with names generated by frugal for struct args, struct return type, and constructors for other structs.

Previously field names were being created in this manner, and referenced by being formatted without that extra rule. As field names aren't a top level name that could conflict with internal frugal, they don't need to be generated in this manner. So I changed all field names to be formatted without that extra rule


aviary-wf:
## Security Insights

## Action Items
---

kevinfrommelt-wf:
+1

blakeroberts-wk:
+1

PR #1002, opened 2018-05-30
Python: Allow generation using absolute output directory path

brandonbodnar-wk:
When given an absolute path for the output directory, the existing SetupGenerator
code for python either went into an infinite loop once it hit the filesystem root
when walking up the directory hierarchy or would silently error once it hit a
folder the running user did not have write permission. Additionally, the existing
algorithm would cause __init__.py to be created in folders unrelated to frugal code
generation simply because the folder existed along the relative path given for output
directory

Update the python generation code to
- Respect errors returned by the SetupGenerator method
- Only generate __init__.py files in the folders that exist between the root output
  directory and the target output directory for the generator. This change also
  addresses the potential infinite loop issue as the path used in the traversal
  algorithm will now be a relative path regardless of user input


aviary3-wk:
## Security Insights

## Action Items
---

kevinfrommelt-wf:
+1

brianshannan-wf:
+1

PR #1007, opened 2018-06-07
INFENG-5834: Allow loose pinning of minor versions of requests and six

charliestrawn-wf:
The frugal python package pins its versions which causes warnings for consumers if the versions aren't exactly resolved. It would be nice to open up those ranges some so that this doesn't occur.

@Workiva/messaging-pp @Workiva/product2 @georgelesica-wf

### DIFF
--- a/lib/python/setup.py
+++ b/lib/python/setup.py
@@ -22,9 +22,9 @@ setup(
     url='http://github.com/Workiva/frugal',
     packages=find_packages(exclude=('frugal.tests', 'frugal.tests.*')),
     install_requires=[
-        "six==1.10.0",
+        "six>=1.10.0,<2",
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Security Insights

## Action Items
---

tylerrinnan-wf:
Think we can do this for minor versions of thrift as well? I think it's up to 0.11.0 now.

brianshannan-wf:
I think opening up to patch would be fine, but thrift has made breaking changes in the past with minor versions as they're under 1.0, so I'm not sure that's great

brianshannan-wf:
+1

danielnovotny-wf:
+1

PR #1018, opened 2018-07-27
INFENG-6058 Dart: Generate hashCode getter for structs.

brianneal-wf:
Generate the `hashCode` getter for Dart.

The hash function was modeled after the existing `__hash__` for Python.


aviary-wf:
## Security Insights

## Action Items
---

brianneal-wf:
@brettkail-wk @danaoredson-wf @samueljackson-wf @danielszopa-wk @scottwilson-wf

compiler/generator/dartlang/generator.go
```diff
@@ -1152,6 +1155,17 @@ func (g *Generator) generateEquals(s *parser.Struct) string {
 	return contents
 }
 
+func (g *Generator) generateHashCode(s *parser.Struct) string {
+	contents := tab + "int get hashCode {\n"
+	contents += tabtab + "var value = 17;\n"
+	for _, field := range s.Fields {
+		contents += fmt.Sprintf(tabtab + "value = (value * 31) ^ %s.hashCode;\n", field.Name)
```
brettkail-wk: Today I learned that `null.hashCode` is `0`.

brianshannan-wf: You need to use `toFieldName(field.Name)` here to get the proper casing. You can see the effect of this in the tests, for example https://github.com/Workiva/frugal/pull/1018/files#diff-81cceb9a9c78455b09e1f78749023018R238

brianneal-wf: Ah, ok. I will do that, but I don't think I understand why. The IDL has "ID", so I thought that would also map to "ID" in Dart. But when I called `toFieldName` it turns into "iD". Why is that?

brianshannan-wf: We do that to make the generated dart code more idiomatic regardless of how the input idl does the casing

brianshannan-wf:
+1

compiler/generator/dartlang/generator.go
```diff
@@ -1152,6 +1155,18 @@ func (g *Generator) generateEquals(s *parser.Struct) string {
 	return contents
 }
 
+func (g *Generator) generateHashCode(s *parser.Struct) string {
+	contents := tab + "int get hashCode {\n"
```
danielszopa-wk: Could you append `@override` above this line?

Seems like `toString()` and `==` should have this as well

samueljackson-wf:
QA +10, verified hashCode is generated structs converted to dart & that hashing is no longer done based on object identity

blakeroberts-wk:
Will integer overflow in dart cause uneven hash code distributions, and is this a reasonable concern?
@brianshannan-wf @kevinfrommelt-wf

danielszopa-wk:
If integer overflow is a significant concern Dart2 will allow for arbitrarily large integers with the `BigInt` class which may solve that problem

brianneal-wf:
@blakeroberts-wk  Not sure this answers your question, but in dart2js (the environment our customers will be using), Dart integers map to Javascript doubles:
https://webdev.dartlang.org/faq#q-how-are-integers-handled-when-compiled-to-javascript

In Dart 2, when running in the VM, integers will be 64-bit and will wrap when they overflow:
https://www.dartlang.org/articles/dart-vm/numeric-computation

kevinfrommelt-wf:
+1

brianshannan-wf:
So what happens if you have a ton of fields? Because it would correspond to a javascript double does this mean it wouldn't wrap and when instead hit infinity at some point and get stuck there?

brianshannan-wf:
Should we do something like https://github.com/google/quiver-dart/blob/cf4607db1b4a9b0220ab79a98e14f201053238d6/lib/src/core/hash.dart? Worth noting this wouldn't be a problem in python because python supports arbitrary precision numbers

brianneal-wf:
I do not know if Javascript doubles wrap or just saturate.
I thought about using quiver's hash function. But that would mean we either force all clients to include it as a dependency and / or just include the code wholesale in frugal somewhere (with the appropriate license verbiage).
Thoughts?

brianneal-wf:
Random googling reveals:
> JavaScript numbers are always stored as double precision floating point numbers, following the international IEEE 754 standard. 
> Integers (numbers without a period or exponent notation) are accurate up to 15 digits.

danielszopa-wk:
I'd like to point out that saying javascript double is incorrect. Double is not a type in javascript, the only numerical type is `Number`. Integers are turned into javascript Numbers which have a valid range of -2^53 to 2^53. 

I found this with a quick google search, showing the potential faults trying to create a larger number than 2^53 in javascript.

```javascript
Math.pow(2, 53) == Math.pow(2, 53) + 1
>> true
```

brianshannan-wf:
If we do use something like the quiver algorithm I was more thinking just duplicating what they have for a couple reasons
1. They only support up to four fields
2. It's a well known algorithm
3. We wouldn't need to include that dependency

I'm not sure if that's necessary or not, but if javascript numbers don't wrap we might see poor hash values from the current implementation with enough fields?

@danielszopa-wk that seems like an unnecessary distinction, because it has the same semantics as an IEEE double

brianneal-wf:
You can hash more than 4 fields with quiver... see the `hashObjects` function, which takes an iterable.

At my previous job we did a lot of floating point stuff in C++ and you'd get a processor trap if something wrapped, but that was on an embedded processor with hardware support for that. I don't know if browsers do anything and we probably shouldn't rely on that, so I would lean towards doing the quiver algorithm, which was probably chosen with dart2js in mind.

I'd rather just use quiver, but I can understand not wanting to force a dependency.

I can work something up by duplicating the code, but we probably want to put that in a utility file or something and then have the Dart classes include that file. That will probably take me a bit to figure out, so if you want to go that way and have any pointers for doing that please let me know.

brianneal-wf:
I'll try to find out if the values will wrap on dart2js first just to get that data.

brianneal-wf:
So this code (run on dartpad) seems to indicate it wraps just fine. I'd guess its the exclusive-or which is setting high order bits which then get interpreted as different IEEE 754 values.

```
void main() {
  var n = 17;
  var prev = n;
  var fields = [
    "some stupid string",
    "another string that may be interesting",
    "cat",
    "dog",
    "485838204059t848gfjdla;j904gfajsfbalkjgrl;4j343jl;ajgakl;",
  ];
  
  var i = 0;
  var total = 0;
  while (prev <= n) {
    ++total;
    prev = n;
    var s = fields[i];
    i = (i + 1) % fields.length;
    n = (31 * n) ^ s.hashCode;
  }
  print("prev = $prev, n = $n, total = $total");
}
```

It prints:

prev = 2246141622, n = 916547723, total = 6

Only 6 times through the loop and it did this.

The same algorithm in Python prints:

prev = 4487371225968977557, n = -141912840066639750192, total = 2

which can be explained by different hashing functions and number representation.

So it seems to work, although we could certainly switch to using quiver's hash which may have a better pedigree. I'm not sure where this algorithm came from (I just copied it from the Python generator).

Thoughts?

brianshannan-wf:
If it works as intended (which it appears to) then I don't have a problem going forward with this approach. +1

blakeroberts-wk:
+1

blakeroberts-wk:
I think a rebase might make Rosie happy?

teresarevious-wf:
Rosie's dependency scan is due to master not being pulled into this PR. The previous Pr pulled in dependencies and they are not being seen here.

brianshannan-wf:
+1

PR #1025, opened 2018-08-03
Golang: Replacing encoding/binary imports with internal.go

natewoods-wf:
Simple improvement to reduce the dependency tree and see what the release cadence of this repository is.

### DIFF
--- a/lib/go/bounded_memory_buffer.go
+++ b/lib/go/bounded_memory_buffer.go
@@ -14,7 +14,6 @@
 package frugal
 
 import (
-	"encoding/binary"
 	"fmt"
 
 	"git.apache.org/thrift.git/lib/go/thrift"
@@ -60,7 +59,7 @@ func (f *TMemoryOutputBuffer) Reset() {
 // Bytes retrieves the framed contents of the buffer.
 func (f *TMemoryOutputBuffer) Bytes() []byte {
 	data := f.TMemoryBuffer.Bytes()
-	binary.BigEndian.PutUint32(data, uint32(len(data)-4))
+	bigEndianPutUint32(data, uint32(len(data)-4))
 	return data
 }
 
--- a/lib/go/framed_transport.go
+++ b/lib/go/framed_transport.go
@@ -16,7 +16,6 @@ package frugal
 import (
 	"bufio"
 	"bytes"
-	"encoding/binary"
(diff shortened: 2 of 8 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

brianshannan-wf:
What is the rationale for doing this? `encoding/binary` is from the standard library, so I don't see the point of replicating it in our code

natewoods-wf:
When using gopherjs on go code, any unnecessary dependencies increase the generated asset size.  This also includes when building go binaries, any dependency adds to the size of the resulting executable. This adds 16 lines of code to replace a dependency that adds more than 825 lines the `encoding/binary` package contains.

Note: this logic can be found in other languages as well: https://github.com/Workiva/frugal/blob/master/lib/dart/lib/src/frugal/internal/headers.dart#L139-L151

brianshannan-wf:
My main concerns with this are:
- -809 lines seems like a very marginal size decrease. It seems like a slippery slope to where size optimization for a general purpose library would stop. Do we replace `encoding/base64`? Do we decide to rewrite `fmt`? Do we stop using third party dependencies all together?
- What do we do if we need additional functions from `encode/binary` in the future? Do we have to find those and copy/paste them into our library as well?
- What are the testing expectations for the code we're moving into our library?

natewoods-wf:
1. These will be addressed in follow up PRs (`fmt`, `gocql` and frugal generated code specifically). Third party dependencies are fine, its just heavy ones are not desired.  Especially when you can get a decrease of 98.06% with no external impact for minimal effort.
2. If we start using a greater subset of `encode/binary` we can always re-evaluate this consolidation and revert if necessary.
3. I was assuming we would trust the original code as tested, but I can add unit tests for this content if desired.

brianshannan-wf:
- I don't think saying we get a 98.06% improvement is really accurate, you'd need to compare the size of everything before and after, not just this single dependency. I don't think replacing fmt is something we want to do, gocql would probably be fine as I think only the generated code needs it for compatibility with the thrift generated code, which we don't need anymore.
- To me, having 809 extra lines, in a general purpose library, is worth not having to make that decision every time we want to do something with that library.

natewoods-wf:
Great Points! I'll close this effort for now.

PR #946, opened 2017-12-04
change version of multidict

brianshannan-wf:
This was somehow breaking asyncio nats with tls. I spent the day trying to figure out how/why and didn't come up with anything, so I decided to just bump the version down.

The problem is somewhere in https://github.com/aio-libs/multidict/compare/v3.1.3...v3.2.0#diff but I wasn't able to nail it down further

@Workiva/messaging-pp

PR #1029, opened 2018-08-27
Upgrade asyncio-nats-client; Loosen python requirements

theisensanders-wf:
Latest version of asyncio-nats-client includes python3.7 support.

Additionally, generic python requirements should be pinned to a major version to pick up minor version changes and give consumers flexibility in which versions they choose to use. I removed the explicit multidict requirement because it seems that at some point this became unnecessary https://github.com/Workiva/frugal/pull/946

Lastly, I got the tests running on Python3.7 locally but didn't upgrade runner image because it's more work than it's worth.

### DIFF
--- a/lib/python/setup.py
+++ b/lib/python/setup.py
@@ -22,19 +22,17 @@ setup(
     url='http://github.com/Workiva/frugal',
     packages=find_packages(exclude=('frugal.tests', 'frugal.tests.*')),
     install_requires=[
-        "six>=1.10.0,<2",
-        "thrift==0.10.0",
-        "requests>=2.12.5,<3",
+        'six>=1.10.0,<2',
+        'thrift==0.10.0',
+        'requests>=2.12.5,<3',
     ],
     extras_require={
-        'tornado': ["nats-client==0.7.2"],
+        'tornado': ['nats-client==0.7.2'],
         'asyncio': [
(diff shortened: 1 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/python/setup.py
```diff
@@ -22,19 +22,17 @@
...
+        'tornado': ['nats-client==0.7.2'],
         'asyncio': [
-            "async-timeout==2.0.1",
-            "asyncio-nats-client==0.7.2",
-            # DO NOT CHANGE without verifying asyncio nats with tls works
-            "multidict==3.1.3",
-            "aiohttp==3.1.2",
+            'aiohttp>=3.0.9,<4',
+            'asyncio-nats-client==0.8.0',
+            'async-timeout>=2.0.1,<4',
```
brianshannan-wf: It looks like aiohttp requires 3.0+ (https://github.com/aio-libs/aiohttp/blob/master/setup.py#L92), it seems like this should match that

theisensanders-wf: But aiohttp 3.0.9 (which I'm stuck on because of a testing issue) requires <3

brianshannan-wf:
+1

charliestrawn-wf:
+1

PR #1033, opened 2018-10-03
INFENG-6382: DART: change generation of dart exceptions to implement Exception

davewalker-wk:
This PR addresses an issue wherein the generator is building exception classes by `extends Error` rather than the preferred `implements Exception`.

### DIFF
--- a/compiler/generator/dartlang/generator.go
+++ b/compiler/generator/dartlang/generator.go
@@ -637,10 +637,11 @@ func (g *Generator) generateStruct(s *parser.Struct) string {
 	}
 
 	contents += fmt.Sprintf("class %s ", s.Name)
+	contents += "implements thrift.TBase"
 	if s.Type == parser.StructTypeException {
-		contents += "extends Error "
+		contents += ", Exception"
 	}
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

brianshannan-wf:
Unfortunately I don't think this is a change we can make without a major version bump, as it would change how users are able to use errors, such as passing them as arguments or catching them. Before errors could be accepted as arguments or caught as the Error superclass, but that wouldn't work with this change.

davewalker-wk:
Correct, as part of the ticket I did make note that this resolution would result in a major version change.

brianshannan-wf:
At this point we aren't prepared to cut a new major version, due to getting it rolled out to all of our internal upstream dependencies.

davewalker-wk:
It's ready for when 3.x is under development.  Just wanted to address/ticket out this for whenever the next major bump were to happen, since this isn't a critical issue by any means.

brianshannan-wf:
Okay, how would you feel about closing this pull request to keep the repo clutter down and linking to your branch in the ticket?

PR #1034, opened 2018-10-04
GO: Simplify Type Read/Write Generation

natewoods-wf:
### Background
Through performance profiling of our application, we found the frugal generated go code to take up around half the size of our transpiled javascript bundle.

### Description
This change inlines the `readFieldXXX` and `writeFieldXXX` method bodies into their respective `Read` and `Write` methods.  We fallback to generating `writeFieldXXX` for `set`s, `map`s and `list`s as abstracting the common logic for these would require more function invocations than the current implementation.  Inlining reduces total code and eliminates a method invocation for every field on reads.  We put the commonly generated three invocations of `writeFieldBegin`, `writeXXX`, and `writeFieldEnd` for the common base types.  This significantly reduces the amount of generated code, but does not reduce the number of method invocations when writing. 

### TL;DR
Changing how we generate Type Read/Writes shaves 3MB of data off our generated gopherjs bundle.

### Reviewers
@brianshannan-wf @Workiva/messaging-pp @Workiva/product2


aviary-wf:
## Security Insights

## Action Items
---

examples/go/gen-go/v1/music/f_store_service.go
```diff
@@ -475,33 +479,15 @@ func (p *StoreBuyAlbumArgs) Read(iprot thrift.TProtocol) error {
 	return nil
 }
 
-func (p *StoreBuyAlbumArgs) ReadField1(iprot thrift.TProtocol) error {
```
brianshannan-wf: So removing this is technically a breaking change. No one should be using it, but from past experience, someone is probably using this method

natewoods-wf: Looks like we are pretty safe: https://github.com/search?q=org%3AWorkiva+ReadField&type=Code

natewoods-wf: Looks like usage across all of github is generated files or thrift specific: https://github.com/search?l=go&q=ReadField1&type=Code

brianshannan-wf: While I do sympathize that these methods shouldn't have been publicly exposed and shouldn't be used anywhere, I think making a breaking change like this isn't something we want to do

brianshannan-wf: If we do need to go forward with this, I would feel much better if this was implemented as an opt-in generation option until we are ready for a 3.0 release

compiler/generator/golang/generator.go
```diff
@@ -1030,6 +1083,11 @@ func (g *Generator) GenerateTypesImports(file *os.File) error {
 	} else {
 		contents += "\t\"git.apache.org/thrift.git/lib/go/thrift\"\n"
 	}
+	if g.Options[frugalImportOption] != "" {
+		contents += "\t\"" + g.Options[frugalImportOption] + "\"\n"
+	} else {
+		contents += "\tfrugal \"github.com/Workiva/frugal/lib/go\"\n"
```
brianshannan-wf: This import doesn't need an alias

compiler/generator/golang/generator.go
```diff
@@ -1106,7 +1164,7 @@ func (g *Generator) GenerateServiceImports(file *os.File, s *parser.Service) err
 	if g.Options[frugalImportOption] != "" {
 		imports += "\t\"" + g.Options[frugalImportOption] + "\"\n"
 	} else {
-		imports += "\t\"github.com/Workiva/frugal/lib/go\"\n"
+		imports += "\tfrugal \"github.com/Workiva/frugal/lib/go\"\n"
```
brianshannan-wf: This import doesn't need an alias

compiler/generator/golang/generator.go
```diff
@@ -2156,6 +2214,33 @@ func (g *Generator) getGoTypeFromThriftTypePtr(t *parser.Type, pointer bool) str
 	}
 }
 
+func (g *Generator) getGoTypeFromThriftTypeEnum(typ *parser.Type) string {
```
brianshannan-wf: I think I recognize this from somewhere else in the code, can this function be used there as well, or possibly combined with the type functions above?

lib/go/encoder.go
```diff
@@ -0,0 +1,129 @@
+package frugal
+
+import "git.apache.org/thrift.git/lib/go/thrift"
+
+// WriteString writes string `value` of field name and id `name` and `field` respectively into `p`.
+func WriteString(p thrift.TProtocol, value, name string, field int16) error {
+	if err := p.WriteFieldBegin(name, thrift.STRING, field); err != nil {
+		return thrift.PrependError("write field begin error: ", err)
```
brianshannan-wf: It would be good to get the field id and name in these error messages, so it's easier to tell what went wrong when errors are received

compiler/generator/golang/generator.go
natewoods-wf: The similar usage is in `generateWriteFieldRec` but also wrap it with the thrift protocol reading methods.  This is just the core go type.

lib/go/encoder.go
natewoods-wf: This responsibility is left to the caller, see these lines of `generateWriteFieldInline`

```go
	contents += indent + fmt.Sprintf("\tif err := frugal.Write%s(oprot, %s, \"%s\", %d); err != nil {\n", writeMethod, structField, field.Name, field.ID)
	contents += indent + fmt.Sprintf("\t\treturn thrift.PrependError(fmt.Sprintf(\"%%T::%s:%d \", p), err)", field.Name, field.ID)
```

compiler/generator/golang/generator.go
brianshannan-wf: Ah, right

compiler/generator/golang/generator.go
```diff
@@ -895,7 +898,57 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, first bool) string
...
+
+func (g *Generator) generateWriteFieldInline(field *parser.Field) (contents string) {
+	if !g.skipStandaloneFieldHandler(field) {
+		return fmt.Sprintf("\tif err := p.writeField%d(oprot); err != nil {\n\t\treturn err\n\t}\n", field.ID)
+	}
+
+	// Check if this field is optional and add nil checks if we need them.
+	var indent string
+	var tail string
+	if field.Modifier == parser.Optional || g.isPointerField(field) {
```
brianshannan-wf: Why include `g.isPointerField()`? In the current code it only has an `IsSet()` check if the field is optional, which I believe is the desired behaviour

brianshannan-wf: https://github.com/Workiva/frugal/blob/master/compiler/generator/golang/generator.go#L904 for reference

natewoods-wf: good catch, cary over from when I was trying to determine if the value was nil manually.

natewoods-wf: With the new `WriteStruct` method, if the pointer isn't checked for nil, we will panic in serialization.  And we generate the `IsSet` check for pointer types too.

https://github.com/Workiva/frugal/blob/master/compiler/generator/golang/generator.go#L525

brianshannan-wf:
+1

benechols-wf:
+1

brianshannan-wf:
+1

PR #1035, opened 2018-10-05
Infeng-6076: Upgrade java-nats client in frugal

yuanmwang-wf:
### Overview 

- Updating jnats to 2.1.1

This should be ready for review now

@Workiva/product2

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -212,7 +212,7 @@ public class FNatsServer implements FServer {
     public void serve() throws TException {
         Dispatcher dispatcher = conn.createDispatcher(newRequestHandler());
         for (String subject : subjects) {
-            if (queue.equals("")) {
+            if (queue != null && !queue.isEmpty()) {
                 dispatcher.subscribe(subject);
             } else {
                 dispatcher.subscribe(subject, queue);
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -212,7 +212,11 @@ public class FNatsServer implements FServer {
     public void serve() throws TException {
         Dispatcher dispatcher = conn.createDispatcher(newRequestHandler());
         for (String subject : subjects) {
-            dispatcher.subscribe(subject, queue);
+            if (queue.equals("")) {
+                dispatcher.subscribe(subject);
+            } else {
+                dispatcher.subscribe(subject, queue);
+            }
         }
 
         LOGGER.info("Frugal server running...");
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -18,16 +18,15 @@ import com.workiva.frugal.protocol.FProtocolFactory;
 import com.workiva.frugal.transport.TMemoryOutputBuffer;
 import com.workiva.frugal.util.BlockingRejectedExecutionHandler;
 import io.nats.client.Connection;
+import io.nats.client.Dispatcher;
 import io.nats.client.MessageHandler;
-import io.nats.client.Subscription;
 import org.apache.thrift.TException;
 import org.apache.thrift.transport.TMemoryInputTransport;
 import org.apache.thrift.transport.TTransport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
-import java.util.ArrayList;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CountDownLatch;
@@ -210,9 +209,9 @@ public class FNatsServer implements FServer {
      */
     @Override
     public void serve() throws TException {
-        ArrayList<Subscription> subscriptionArrayList = new ArrayList<>();
+        Dispatcher dispatcher = conn.createDispatcher(newRequestHandler());
         for (String subject : subjects) {
-            subscriptionArrayList.add(conn.subscribe(subject, queue, newRequestHandler()));
+            dispatcher.subscribe(subject, queue);
         }
 
         LOGGER.info("Frugal server running...");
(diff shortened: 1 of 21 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -96,7 +96,7 @@ private FNatsServer(Connection conn, FProcessor processor, FProtocolFactory prot
         private final FProtocolFactory protoFactory;
         private final String[] subjects;
 
-        private String queue = "";
+        private String queue = UUID.randomUUID().toString().replace("-", "");
```
brianshannan-wf: We should leave the default as an empty string. I think using a uuid should do the same thing as an empty string, but an empty string has clearer intent to me

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -64,13 +65,12 @@ protected FNatsSubscriberTransport(Connection conn, String queue) {
...
         /**
-         * Creates a NATS FSubscriberTransportFactory using the provided NATS connection. Subscribers using this
-         * transport will not use a queue.
+         * Creates a NATS FSubscriberTransportFactory using the provided NATS connection.
          *
          * @param conn NATS connection
          */
         public Factory(Connection conn) {
-            this(conn, null);
+            this(conn, UUID.randomUUID().toString().replace("-", ""));
```
brianshannan-wf: Same comment about default queue groups

lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
```diff
@@ -39,12 +40,12 @@
...
     public static final int NATS_MAX_MESSAGE_SIZE = 1024 * 1024;
     public static final String FRUGAL_PREFIX = "frugal.";
+    private static final String INBOX_PREFIX = "_INBOX.";
 
     private final Connection conn;
     private final String subject;
     private final String inbox;
-
-    private Subscription sub;
+    private Dispatcher dispathcer;
```
brianshannan-wf: typo

yuanmwang-wf:
@brianshannan-wf I remember now, setting queue name to empty string is a no-op now:

```
java.lang.IllegalArgumentException: QueueName is required in subscribe
	at io.nats.client.impl.NatsDispatcher.subscribe(NatsDispatcher.java:166)
	at com.workiva.frugal.transport.FNatsSubscriberTransport.subscribe(FNatsSubscriberTransport.java:126)
	at frugal.test.EventsSubscriber$Client.subscribeEventCreated(EventsSubscriber.java:87)
	at com.workiva.TestClient.main(TestClient.java:628)
```
https://wf-skynet-hrd.appspot.com/results/smithy/1492195/1#results

Guess we should change it back to UUIDs?

brianshannan-wf:
@yuanmwang-wf It's an overloaded function, we should use https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/NatsDispatcher.java#L152 if a queue name isn't needed

yuanmwang-wf:
@brianshannan-wf Ahh yeah you're right, these should be addressed now.

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -125,21 +123,26 @@ public void subscribe(String topic, FAsyncCallback callback) throws TException {
             } catch (TException ignored) {
             }
         });
+        if (queue.equals("")) {
```
brianshannan-wf: check for null too?

yuanmwang-wf: addressed in 5bd720a

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -222,11 +226,11 @@ public void serve() throws TException {
         }
         LOGGER.info("Frugal server stopping...");
 
-        for (Subscription subscription : subscriptionArrayList) {
+        for (String subject : subjects) {
```
brianshannan-wf: We should close the dispatcher as well. https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/Connection.java#L217. According to the docs that will unsubscribe everything as well, but we should verify that's the case if we go with that

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -222,11 +226,11 @@ public void serve() throws TException {
...
             try {
-                subscription.unsubscribe();
-            } catch (IOException e) {
-                LOGGER.warn("Frugal server failed to unsubscribe from " + subscription.getSubject() + ": " +
+                dispatcher.unsubscribe(subject);
+            } catch (IllegalArgumentException e) {
```
brianshannan-wf: It doesn't look like this can throw an IllegalArgumentException, rather and IllegalStateException. https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/NatsDispatcher.java#L201

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -98,22 +97,21 @@ public FNatsSubscriberTransport getTransport() {
 
     @Override
     public boolean isSubscribed() {
-        return conn.getState() == Nats.ConnState.CONNECTED && sub != null && sub.isValid();
+        return conn.getStatus() == Status.CONNECTED && dispatcher.isActive();
```
brianshannan-wf: I would feel better if we check is dispatcher is null here like we did with sub

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -125,21 +123,26 @@ public void subscribe(String topic, FAsyncCallback callback) throws TException {
...
-        if (sub == null) {
+        if (dispatcher == null) {
             LOGGER.warn("attempted to unsubscribe with a null internal " +
                     "subscription - possibly unsubscribing more than once - subject: " + subject);
             return;
         }
         try {
-            sub.unsubscribe();
-        } catch (IOException e) {
+            dispatcher.unsubscribe(getFormattedSubject());
```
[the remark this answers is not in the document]
yuanmwang-wf: What do you mean by "close the dispatcher"? It should be set to null in line 145, is there something else needed when unsubscribing?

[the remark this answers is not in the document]
yuanmwang-wf: Umm nvm.... For some reason I didn't see your previous comment about closing the dispatcher (facepalm). I've added close in both places - I've kept the unsub calls in frugal instead of relying on `closeDispatcher` to unsubscribe everything, just thought it looks clearer.

brianshannan-wf:
+1 on the code. Would like to have a little QA done before we merge this

brianshannan-wf:
+10

blakeroberts-wk:
+1

brianshannan-wf:
Before this is merged I'd like to have the messaging-sdk pr as done as possible so we don't have a large amount of time with this dependency mismatch

PR #1039, opened 2018-10-26
Dart: eliminate the map of FMethods

toddbeckman-wf:
# Description

Part of PLC-9

The use of a map of `FMethod` object makes it impossible for tree-shaking algorithms to optimize any unused methods (which there are plenty of, due to our fix-forward practice).

- This PR removes the map in Dart. We can simply compose the middleware on the fly. This saves memory, extra allocations, and solves the problem above.
- And because we can compose middleware on the fly, we can eliminate all usages of `FMethod` and deprecated it.

# TODO

- Make these same changes in Go (again, could come in a follow-up)


aviary-wf:
## Security Insights

## Action Items
---

lib/dart/lib/src/frugal/f_middleware.dart
```diff
@@ -49,19 +65,9 @@ class FMethod {
...
-
-    if (middleware == null) {
-      return handler;
-    }
-    // ignore: STRONG_MODE_DOWN_CAST_COMPOSITE
-    return middleware.fold(handler, (prev, element) => element(prev));
-  }
+  InvocationHandler _composeMiddleware(
+          dynamic f, List<Middleware> middleware) =>
+      composeMiddleware(f, middleware);
```
toddbeckman-wf: dartfmt caused this monstrosity

lib/dart/lib/src/frugal/f_middleware.dart
```diff
@@ -26,8 +26,24 @@ typedef Future InvocationHandler(
...
+  if (middleware == null) {
+    return handler;
+  }
+  // ignore: STRONG_MODE_DOWN_CAST_COMPOSITE
+  return middleware.fold(handler, (prev, element) => element(prev));
+}
+
 /// Contains an [InvocationHandler] used to proxy the given service method
 /// This should only be used by generated code.
+@deprecated
```
toddbeckman-wf: Now unused in all generated Dart code :)

compiler/generator/dartlang/generator.go
```diff
@@ -1398,7 +1393,8 @@ func (g *Generator) GeneratePublisher(file *os.File, scope *parser.Scope) error
 
 		publishers += fmt.Sprintf(tab+"Future publish%s(frugal.FContext ctx, %s%s req) {\n", op.Name, args, g.getDartTypeFromThriftType(op.Type))
 
-		publishers += fmt.Sprintf(tabtab+"return this._methods['%s']([ctx, %sreq]);\n", op.Name, argsWithoutTypes)
+		publishers += fmt.Sprintf(tabtab+"return frugal.composeMiddleware(_publish%s, _combinedMiddleware)('%s', 'publish%s', [ctx, %sreq]);\n",
```
brianshannan-wf: Won't this call `composeMiddleware()` for every function invocation? Before the middleware only needed to be composed once per method because it happened in the constructor for `FMethod` when the generated code object was constructed

lib/dart/lib/src/frugal/f_middleware.dart
brianshannan-wf: Assuming this can be made public, we should remove the private method and having things just reference the public one

compiler/generator/dartlang/generator.go
toddbeckman-wf: What do you think instead of the first commit's solution of using private fields instead: https://github.com/Workiva/frugal/pull/1039/commits/9b1b844d4cb001ee660821e5ad46897e68ff3990

brianshannan-wf: Here's my thought process on that. Caching the composed middleware is better than doing it every time, but as this is a core piece of how frugal works, I'd like to keep a similar implementation across languages as it makes maintenance and understanding how everything works much easier. Which then means you'd have to do locking around that for every method call in other languages, which seems problematic. I'm not sure if there's a way to compose the middleware and pass a function to the composed middleware after the fact easily or something else

toddbeckman-wf:
Reviewers: note the conversation in an oudated diff about the use of class fields.

stevenosborne-wf:
Due to concerns around dart deviating too far from the other languages, closing for now. Can revisit in the future.

PR #1040, opened 2018-10-26
PLC-9 Dart: remove extra _isset_X fields

toddbeckman-wf:
# Description

Part of PLC-9

These are unnecessary as Dart allows nullable primitives. Removing them essentially halves the number of fields both in the code and in memory of any structs that use primitives.

# Testing

Install this branch and build a large consumer project. Ensure the Dart code is syntatically correct and nothing is broken.

fyi 


compiler/generator/dartlang/generator.go
```diff
@@ -654,19 +654,12 @@ func (g *Generator) generateStruct(s *parser.Struct) string {
 	// Fields
 	for _, field := range s.Fields {
 		contents += g.generateCommentWithDeprecated(field.Comment, tab, field.Annotations)
-		contents += fmt.Sprintf(tab+"%s _%s%s;\n",
+		contents += fmt.Sprintf(tab+"%s %s%s;\n",
```
toddbeckman-wf: Since the getter is just getting this and the setter is just setting this, we can just use a public field for identical functionality and API. This eliminates unneeded getters and setters from all fields.

compiler/generator/dartlang/generator.go
```diff
@@ -654,19 +654,12 @@ func (g *Generator) generateStruct(s *parser.Struct) string {
...
 			g.getDartTypeFromThriftType(field.Type), toFieldName(field.Name), g.generateInitValue(field))
 		contents += fmt.Sprintf(tab+"static const int %s = %d;\n", strings.ToUpper(field.Name), field.ID)
 	}
 	contents += "\n"
 
-	// Is set helpers for primitive types.
-	for _, field := range s.Fields {
-		if g.isDartPrimitive(field.Type) {
-			contents += fmt.Sprintf(tab+"bool __isset_%s = false;\n", toFieldName(field.Name))
```
toddbeckman-wf: We just check if `field == null`, so the field-specific `__isset_%s` is entirely unneeded.

compiler/generator/dartlang/generator.go
```diff
@@ -725,42 +718,8 @@ func (g *Generator) generateInitValue(field *parser.Field) string {
 }
 
 func (g *Generator) generateFieldMethods(s *parser.Struct) string {
-	// Getters and setters for each field
```
toddbeckman-wf: :)

compiler/generator/dartlang/generator.go
```diff
@@ -725,38 +718,11 @@ func (g *Generator) generateInitValue(field *parser.Field) string {
...
-			contents += fmt.Sprintf(tab+"unset%s() {\n", titleName)
-			contents += fmt.Sprintf(tabtab+"this.__isset_%s = false;\n", fName)
-			contents += tab + "}\n\n"
-		} else {
-			contents += fmt.Sprintf(tab+"bool isSet%s() => this.%s != null;\n\n", titleName, fName)
-			contents += fmt.Sprintf(tab+"unset%s() {\n", titleName)
-			contents += fmt.Sprintf(tabtab+"this.%s = null;\n", fName)
-			contents += tab + "}\n\n"
-		}
+		contents += fmt.Sprintf(tab + "@deprecated\n")
```
toddbeckman-wf: These are unneeded, but public API, so deprecate for now.

compiler/generator/dartlang/generator.go
```diff
@@ -777,11 +743,11 @@ func (g *Generator) generateFieldMethods(s *parser.Struct) string {
...
 		fName := toFieldName(field.Name)
 		contents += fmt.Sprintf(tabtabtab+"case %s:\n", strings.ToUpper(field.Name))
-		contents += tabtabtabtab + "if(value == null) {\n"
-		contents += fmt.Sprintf(tabtabtabtabtab+"unset%s();\n", strings.Title(field.Name))
-		contents += tabtabtabtab + "} else {\n"
-		contents += fmt.Sprintf(tabtabtabtabtab+"this.%s = value as %s;\n", fName, g.getDartTypeFromThriftType(field.Type))
-		contents += tabtabtabtab + "}\n"
+		contents += tabtabtabtab + "if (value == null) {\n"
+		contents += fmt.Sprintf(tabtabtabtabtab+"%s = null;\n", fName)
+		contents += fmt.Sprintf(tabtabtabtab+"} else if (value is %s) {\n", g.getDartTypeFromThriftType(field.Type))
```
toddbeckman-wf: Dart prefers `is to `as`.

compiler/generator/dartlang/generator.go
```diff
@@ -884,20 +856,13 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, first bool, ind st
 		}
 
 		contents += fmt.Sprintf(ind+"%s%s = iprot.read%s();\n", prefix, fName, thriftType)
-		if primitive && first {
-			contents += fmt.Sprintf(ind+"this.__isset_%s = true;\n", fName)
```
toddbeckman-wf: Assigning values here at all replaces the need for this isset variable.

compiler/generator/dartlang/generator.go
```diff
@@ -884,20 +856,13 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, first bool, ind st
...
 	} else if g.Frugal.IsEnum(underlyingType) {
 		if g.useEnums() {
 			contents += fmt.Sprintf(ind+"%s%s = %s.deserialize%s(iprot.readI32());\n",
 				prefix, fName, g.includeQualifier(underlyingType), underlyingType.Name)
 		} else {
 			contents += fmt.Sprintf(ind+"%s%s = iprot.readI32();\n", prefix, fName)
 		}
-
-		if first {
-			contents += fmt.Sprintf(ind+"this.__isset_%s = true;\n", fName)
```
toddbeckman-wf: Assigning values here at all replaces the need for this isset variable.

compiler/generator/dartlang/generator.go
```diff
@@ -1198,9 +1163,13 @@ func (g *Generator) generateValidate(s *parser.Struct) string {
 	contents := tab + "validate() {\n"
 
 	if s.Type != parser.StructTypeUnion {
-		contents += tabtab + "// check for required fields\n"
+		first := true
 		for _, field := range s.Fields {
 			if field.Modifier == parser.Required {
+				if first {
+					first = false
+					contents += tabtab + "// check for required fields\n"
```
toddbeckman-wf: Just some #nits: this and the similar comment below were getting added regardless of whether the checks were being added, so now only add them if we're checking.

compiler/generator/dartlang/generator.go
brianshannan-wf: This isn't true. If a field has a default value, it's not considered "set" if it's left at the default value

brianshannan-wf: We may be able to remove the need for a field to track this, but I'm fairly certain we still need a method at least

toddbeckman-wf: I've updated the check. What do you think of this approach?

examples/dart/gen-dart/v1_music/lib/src/f_store_service.dart
```diff
@@ -455,12 +435,12 @@ class buyAlbum_result implements thrift.TBase {
     validate();
 
     oprot.writeStructBegin(_STRUCT_DESC);
-    if (isSetSuccess() && this.success != null) {
+    if (success != null && this.success != null) {
```
brettkail-wk: This is now obviously redundant.

examples/dart/gen-dart/v1_music/lib/src/f_store_service.dart
```diff
@@ -473,7 +453,7 @@ class buyAlbum_result implements thrift.TBase {
   String toString() {
     StringBuffer ret = new StringBuffer("buyAlbum_result(");
 
-    if (isSetSuccess()) {
+    if(success != null) {
```
brettkail-wk: This looks misformatted.

examples/dart/gen-dart/v1_music/lib/src/f_store_service.dart
```diff
@@ -584,18 +550,20 @@ class enterAlbumGiveaway_args implements thrift.TBase {
     switch (fieldID) {
       case EMAIL:
         if (value == null) {
-          unsetEmail();
-        } else {
-          this.email = value as String;
+          email = null;
+        } else if (value is String) {
+          email = value;
```
brettkail-wk: This looks wrong: previously if `value` was not a string then it failed with an error.  Now it silently ignores the invalid value.

examples/dart/gen-dart/v1_music/lib/src/f_track.dart
```diff
@@ -402,9 +371,7 @@ class Track implements thrift.TBase {
   }
 
   validate() {
-    // check for required fields
-    // check that fields of type enum have valid values
-    if (isSetPro() && !t_v1_music.PerfRightsOrg.VALID_VALUES.contains(pro)) {
+    if(pro != null && !t_v1_music.PerfRightsOrg.VALID_VALUES.contains(pro)) {
```
brettkail-wk: This looks misformatted.

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
```diff
@@ -13,54 +13,34 @@ class Album implements thrift.TBase {
   static final thrift.TField _DURATION_FIELD_DESC = new thrift.TField("duration", thrift.TType.DOUBLE, 2);
   static final thrift.TField _ASIN_FIELD_DESC = new thrift.TField("ASIN", thrift.TType.STRING, 3);
 
-  List<t_v1_music.Track> _tracks;
+  List<t_v1_music.Track> tracks;
```
brettkail-wk: I don't know Dart well enough.  Is there any risk that someone subclassed a Frugal class and overrode a getter/setter?  Is that supported, and if so, do we care?

examples/dart/gen-dart/v1_music/lib/src/f_store_service.dart
toddbeckman-wf: :facepalm:

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
toddbeckman-wf: There is already a getter and setter for this field, which in Dart is equivalent to just making this a public field.

examples/dart/gen-dart/v1_music/lib/src/f_store_service.dart
toddbeckman-wf: Addressed.

examples/dart/gen-dart/v1_music/lib/src/f_purchasing_error.dart
```diff
@@ -64,18 +50,20 @@ class PurchasingError extends Error implements thrift.TBase {
     switch (fieldID) {
       case MESSAGE:
         if (value == null) {
-          unsetMessage();
+          message = null;
         } else {
-          this.message = value as String;
+          message = value as String; // ignore: avoid_as
```
brettkail-wk: Does `value as String` work for null?  i.e., can the `if (value == null)` be removed altogether leaving simply `message = value as String`?

blakeroberts-wk: Yes,
```dart
assert(null as String == null); // passes

// note: null can be cast but doesn't get the type
assert((null as String) is String); // fails (but doesn't matter in this case)
```

brettkail-wk: Then I'd suggest further simplifying the generated code.

toddbeckman-wf: Updated

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
```diff
@@ -188,7 +157,7 @@ class Album implements thrift.TBase {
     oprot.writeFieldBegin(_DURATION_FIELD_DESC);
     oprot.writeDouble(duration);
     oprot.writeFieldEnd();
-    if (this.aSIN != null) {
+    if (ASIN != null) {
```
brettkail-wk: Some of these changes to `write` methods look wrong (some but not all are using upper-cased).

What is the testing strategy for this PR?  It seems we should run this against a large internal repo and make sure Dart clients are still working as expected.

compiler/generator/dartlang/generator.go
brianshannan-wf: I guess I don't understand why this should be deprecated if it's checking against a default value as well

compiler/generator/dartlang/generator.go
```diff
@@ -1272,13 +1242,16 @@ func (g *Generator) generateValidate(s *parser.Struct) string {
 	}
 
 	if !g.useEnums() {
-		contents += tabtab + "// check that fields of type enum have valid values\n"
+		first := false
```
brianshannan-wf: `first := true`?

toddbeckman-wf: Good find. However, all this affects is whether a comment gets added. Are these comments necessary or could I just rip out the `first` check here and elsewhere?

brianshannan-wf:
+1

tylerrinnan-wf:
+1

stevenosborne-wf:
@Workiva/release-management-p

patkujawa-wf:
This seems to have a bug. At least for `i64` and maybe other primitives, the generated code provides a default value (e.g. `0`) (even if no default value is provided in the frugal IDL), which makes a null-check incorrect. This was @brianshannan-wf 's concern, mentioned in a comment above. Here's the transcript from support chat:

pat.kujawa [12:36 PM]
:hieverybody: @nathan.coleman and I think a recent frugal version has a bug. [PLC-9 Dart: remove extra _isset_X fields by toddbeckman-wf · Pull Request #1040 · Workiva/frugal](https://github.com/Workiva/frugal/pull/1040) might have introduced it. It's causing two tests to fail: https://wf-skynet-hrd.appspot.com/integration-suite-results/ag9zfndmLXNreW5ldC1ocmRyOAsSFEludGVncmF0aW9uVGVzdFN1aXRlIh5zaGlweWFyZDpza2FhcmRiOmRhcnQ6NTE4Njg0LjEM?filter=Failed

[`expect(openFragRequest.isSetSectionID(), isFalse);`](https://github.com/Workiva/skaardb/blob/e8430c5273acd4f7df977c550008a9301e95efae/test/unit/doc_plat_client/doc_plat_client_test.dart#L3105) is the code that's now failing. Looking at the generated code, `bool isSetSectionID() => sectionID != null;` is the new implementation. Which would be correct, except that field is initialized to `int sectionID = 0;`.

nathan.coleman [12:37 PM]
We've isolated the problem by updating the Frugal dependency in skaardb here: https://github.com/nathancoleman-wf/skaardb/compare/master...nathancoleman-wf:update_frugal_msdk

pat.kujawa [12:40 PM]
The frugal IDL doesn't specify a default value (just [`4: i64 sectionID,`](https://github.com/Workiva/skaardb/blob/695758356de596061da2572db7cf6820e1135185/frontend/messaging/idl/api/v3/fragments.frugal#L47)), and the old frugal generator also initialized the value to `0`, so it seems isolated to the new isset logic.
It's quite an insidious bug, too, because the types are the same so you have to have tests in place to catch it

nathan.coleman [12:42 PM]
We should note that both failing tests are due to `.isSetSectionID()` returning `true` when `false` is expected

pat.kujawa [12:42 PM]
https://github.com/nathancoleman-wf/skaardb/compare/master...nathancoleman-wf:update_frugal_msdk#diff-5a429de4fe07110bbc8a4e8bc2c11f1e shows the diff of the generated dart code, for reference.

yuan.m.wang [12:44 PM]
I agree, that sounds like a bug in frugal
I can start working on a fix now

PR #1042, opened 2018-11-08
INFENG-6428 INFENG-6429 RELEASE 2.24.0

brianshannan-wf:


aviary3-wk:
## Security Insights

## Action Items
---

brianshannan-wf:
So I didn't think about this before, but is the java update actually a breaking change for users of frugal? 
Is the connection in https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java#L67 technically a different class?

brianshannan-wf:
I'm fairly certain this is the case, which means we probably have to wait until 3.0 for this change unless I'm missing something?

PR #1047, opened 2018-11-13
Infeng-6366: amazon mq frugal transport

yuanmwang-wf:
**Overview**
Implements go mq transport in frugal.

I've created a new branch `amazon-mq` so all mq transport PRs can go into this branch and test together. 

@Workiva/product2


lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+	"sync"
+
+	"git.apache.org/thrift.git/lib/go/thrift"
+	"github.com/go-stomp/stomp"
+)
+
+const frugalVirtualTopicPrefix = "frugal.VirtualTopic."
+
+// FAmazonMqPublisherTransportFactory creates FMqPublisherTransports.
+type FAmazonMqPublisherTransportFactory struct {
```
brianshannan-wf: IMO we should name this something related to stomp, as this transport should work for all message brokers that support the stomp protocol

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// IsOpen returns true if the transport is open, false otherwise.
+func (m *fAmazonMqPublisherTransport) IsOpen() bool {
+	return m.conn != nil
+}
+
+// Close closes the transport.
+func (m *fAmazonMqPublisherTransport) Close() error {
+	if err := m.conn.Disconnect(); err != nil {
```
brianshannan-wf: If we aren't opening the connection in open, we shouldn't close it in close. For the nats transport we decided the connection lifecycle should be managed outside the transport, so there is precedent for that

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	m.conn = nil
+	return nil
+}
+
+// GetPublishSizeLimit returns the maximum allowable size of a payload
+// to be published. A non-positive number is returned to indicate an
+// unbounded allowable size.
+func (m *fAmazonMqPublisherTransport) GetPublishSizeLimit() uint {
+	return 0
```
brianshannan-wf: I think I read the default max message size was 32 mb for activemq, not sure if we have that configured to something else in our broker.

brianshannan-wf: Additionally, maybe making this configurable would be good for future proofing? Might be a case of wait until you need it to be configurable though

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+import (
+	"bytes"
+	"fmt"
+	"sync"
+
+	"git.apache.org/thrift.git/lib/go/thrift"
+	"github.com/go-stomp/stomp"
+)
+
+const frugalVirtualTopicPrefix = "frugal.VirtualTopic."
```
brianshannan-wf: IMO we want the `VirtualTopic.` part of this to be passed in, for two reasons
1. People in the future may want to not use virtual topics if they don't want those semantics, and I think making people ask for virtual topics is better than assuming they want them
2. Other brokers than activemq might not have the same semantics or naming system

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+
+// Publish sends the given payload with the transport. Data are sent to virtual topics created in AmazonMq.
+func (m *fAmazonMqPublisherTransport) Publish(topic string, data []byte) error {
+	if !m.IsOpen() {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_NOT_OPEN, "frugal: mq transport not open")
+	}
+
+	destination := fmt.Sprintf("/topic/%s%s", frugalVirtualTopicPrefix, topic)
+
+	if err := m.conn.Send(destination, "text/plain", data); err != nil {
```
brianshannan-wf: We're sending arbitrary binary data, doesn't seem like `text/plain` is the proper encoding. Should probably be a binary encoding if that exists. If not, we'll probably have to base64 encode the data, which isn't super desirable because then subscribers in other languages will have to know to expect that

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if m.isSubscribed {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_ALREADY_OPEN, "frugal: mq transport already has a subscription")
+	}
+
+	if queue == "" {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_UNKNOWN, "frugal: mq transport cannot subscribe to empty queue")
+	}
+
+	destination := fmt.Sprintf("/queue/frugalConsumer.%s.%s", m.consumerName, queue)
```
brianshannan-wf: Same comment about topic name shouldn't be for virtual topics by default

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// NewAmazonMqFSubscriberTransport creates a new FSubscriberTransport which is used for
+// pub/sub with a queue.
+func NewAmazonMqFSubscriberTransportWithQueue(conn *stomp.Conn, consumerName string, queue string) FSubscriberTransport {
+	return &fAmazonMqSubscriberTransport{conn: conn, consumerName: consumerName, queue: queue}
+}
+
+// Subscribe sets the subscribe queue and opens the transport.
+func (m *fAmazonMqSubscriberTransport) Subscribe(queue string, callback FAsyncCallback) error {
```
brianshannan-wf: s/queue/topic ? (or subject, we've used the term pretty interchangeably in the past)

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+	if !m.isSubscribed {
+		return nil
+	}
+
+	m.stopC <- true
+	if err := m.sub.Unsubscribe(); err != nil {
+		return thrift.NewTTransportExceptionFromError(err)
+	}
+
+	if err := m.conn.Disconnect(); err != nil {
```
brianshannan-wf: Same comment about connection disconnection as the publisher transport

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+
+// Unsubscribe unsubscribes and disconnects the stomp connection.
+func (m *fAmazonMqSubscriberTransport) Unsubscribe() error {
+	m.openMu.Lock()
+	defer m.openMu.Unlock()
+	if !m.isSubscribed {
+		return nil
+	}
+
+	m.stopC <- true
```
brianshannan-wf: You could also close the channel instead of putting something on it

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+	m.isSubscribed = false
+	m.callback = nil
+	return nil
+}
+
+// Processes messages from subscription channel with the given FAsyncCallback.
+func (m *fAmazonMqSubscriberTransport) processMessages() {
+	for {
+		select {
+		case <-m.stopC:
```
brianshannan-wf: I think we should grab this channel as a variable before the loop to prevent the following race condition (I think is possible)
1. Subscription created, processMessages() started
2. Unsubscribe called, stop channel indicates loop should close
3. Subscribe called again, stop channel is recreated before this loop can notice the stop channel told it it should shut down, and it sees the new channel in this loop. Leading to two loops processing messages

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+			return
+		case message, ok := <-m.sub.C:
+			if !ok {
+				logger().Errorf("messaging_sdk: error processing subscription messages, message channel closed unexpectedly")
```
brianshannan-wf: s/messaging_sdk/frugal

lib/go/amazon_mq_transport.go
```diff
@@ -0,0 +1,236 @@
...
+				return
+			}
+
+			if len(message.Body) < 4 {
+				continue
+			}
+
+			transport := &thrift.TMemoryBuffer{Buffer: bytes.NewBuffer(message.Body[4:])}
+			if m.callback(transport) != nil {
+				continue
```
brianshannan-wf: I think we should log something here if the users callback returned an error

lib/go/amazon_mq_transport.go
yuanmwang-wf: Good point, addressed now.

lib/go/amazon_mq_transport.go
yuanmwang-wf: I've made it to be configurable now 😄

lib/go/amazon_mq_transport.go
yuanmwang-wf: Agreed, changed now the transport will only added a frugal prefix to the topic, so it's up to the client to decide if it wants to use virtual topic or not.

lib/go/amazon_mq_transport.go
yuanmwang-wf: Looks like stomp allows most common MIME types, and based on this [stackoverflow question](https://stackoverflow.com/questions/6783921/which-mime-type-to-use-for-a-binary-file-thats-specific-to-my-program) for arbitrary binary data content-type should be set to `application/octet-stream`.

lib/go/amazon_mq_transport.go
yuanmwang-wf: 👍

lib/go/amazon_mq_transport.go
yuanmwang-wf: I think this should be queue? Looks like when producers publish to virtual topics, the broker then makes the messages available in queues.

>The idea behind virtual topics is that producers send to a topic in the usual JMS way. Consumers can continue to use the Topic semantics in the JMS specification. However if the topic is virtual, consumer can consume from a physical queue for a logical topic subscription, allowing many consumers to be running on many machines & threads to load balance the load.

>E.g., let's say we have a topic called VirtualTopic.Orders. (Where the prefix VirtualTopic. indicates its a virtual topic). And we logically want to send orders to systems A and B. Now with regular durable topics we'd create a JMS consumer for clientID_A and "A" along with clientID_B and "B".

>With virtual topics we can just go right ahead and consume to queue Consumer.A.VirtualTopic.Orders to be a consumer for system A or consume to Consumer.B.VirtualTopic.Orders to be a consumer for system B.

lib/go/amazon_mq_transport.go
yuanmwang-wf: 👍

lib/go/amazon_mq_transport.go
yuanmwang-wf: Should be addressed now.

lib/go/amazon_mq_transport_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/go-stomp/stomp/server"
+	"github.com/stretchr/testify/assert"
+)
+
+const stompPort = ":61614"
+
+// Ensures Amazon Mq transport is able to open and close.
+func TestAmazonMqPublisherOpenPublish(t *testing.T) {
+	// starts a tcp server.
+	l, _ := net.Listen("tcp", stompPort)
```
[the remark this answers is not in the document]
yuanmwang-wf: I've noticed the unit tests are indeed intermittently failing in workiva build because of this. I have not found any obvious interface that could be mocked, I took the test pattern from [here](https://github.com/go-stomp/stomp/blob/master/server/server_test.go#L48), I'll keep looking.

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,236 @@
...
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_NOT_OPEN, "frugal: stomp transport not open")
+	}
+
+	if len(data) > m.maxPublishSize {
+		return thrift.NewTTransportException(
+			TRANSPORT_EXCEPTION_REQUEST_TOO_LARGE,
+			fmt.Sprintf("Message exceeds %d bytes, was %d bytes", m.maxPublishSize, len(data)))
+	}
+
+	destination := fmt.Sprintf("/topic/%s%s", frugalPrefix, topic)
```
brianshannan-wf: Don't we need to allow users to pass in a topic prefix to insert here?

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,236 @@
...
+// FStompSubscribeTransportFactory creates fStompSubscriberTransports.
+type FStompSubscribeTransportFactory struct {
+	conn         *stomp.Conn
+	queue        string
+	consumerName string
+}
+
+// NewFStompSubscriberTransportFactory creates FStompSubscribeTransportFactory with the given stomp
+// connection and consumer name.
+func NewFStompSubscriberTransportFactory(conn *stomp.Conn, consumerName string) *FStompSubscribeTransportFactory {
```
brianshannan-wf: Should `consumerName` be named something else? As to use virtual topics a user would need to pass in `Consumer.<consumerName>.VirtualTopic.`

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if m.isSubscribed {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_ALREADY_OPEN, "frugal: stomp transport already has a subscription")
+	}
+
+	if queue == "" {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_UNKNOWN, "frugal: stomp transport cannot subscribe to empty queue")
+	}
+
+	destination := fmt.Sprintf("/queue/%s.%s", m.consumerName, queue)
```
brianshannan-wf: Should this be `/topic`? I know consuming from a virtual topic is actually consuming from a queue under the hood, but I was under the impression activemq abstracted that detail away, and from a protocol point of view was like consuming from a topic. Is that not true?

brianshannan-wf: Additionally, IMO we should make people pass in the `.` as part of the prefix, as if they don't pass in a prefix you end up with a period at the beginning of the topic

lib/go/amazon_mq_transport.go
brianshannan-wf: You're using the topic to subscribe to the virtual topic, and you end up subscribing to a queue under the hood. As this parameter is the topic you're subscribing to and not an object representing the queue we're consuming from, I think `topic` or `subject` better describes what it is

lib/go/amazon_mq_transport_test.go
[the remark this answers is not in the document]
brianshannan-wf: A common approach in go is to make an interface with all the methods you need from a certain struct and use that interface as the type of the field instead of the struct. You can then mock out the interface in unit tests

lib/go/stomp_transport.go
yuanmwang-wf: >Should this be /topic? I know consuming from a virtual topic is actually consuming from a queue under the hood, but I was under the impression activemq abstracted that detail away, and from a protocol point of view was like consuming from a topic. Is that not true?

I'm not sure if jms does the conversion under the hood, but I've tested with go stomp that it actually needs to be reading from the queue instead of topic. The following code publishes to a virtual topic and reads a message from a queue successfully. 

```go
err = stompConn.Send("/topic/VirtualTopic.Orders", "text/plain", []byte(sampleMsg))

sub, err := stompConn.Subscribe("/queue/Consumer.A.VirtualTopic.Orders", stomp.AckAuto)
if err != nil {
	log.Fatalln(err)
}
msg := <-sub.C
log.Println(string(msg.Body[:]))
```

In the broker console the virtual topic is listed under topics: 


But the subscribers are listed under queues:


I've also confirmed if the code just publishes to the virtual topic, the messages enqueued will increment for the queue. So I think it does need to subscribe to the queue instead.

lib/go/amazon_mq_transport.go
yuanmwang-wf: Sounds good 👍

lib/go/stomp_transport.go
yuanmwang-wf: `.` has been removed now (consumers will need to pass formatted string in).

lib/go/stomp_transport.go
yuanmwang-wf: Changed to `consumerPrefix`.

lib/go/stomp_transport.go
yuanmwang-wf: Added `topicPrefix`.

lib/go/amazon_mq_transport_test.go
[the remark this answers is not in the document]
yuanmwang-wf: Actually I just changed the tcp port number in unit tests from a fixed number to `:0` so the test suite will pick a random available port, that seems to have solved the intermittent failures - I've reran smithy 3 times on f0e2456 and all of them passed.

lib/go/stomp_transport.go
brianshannan-wf: Ah yeah good call, I misread the java docs. We should probably have that be configurable then, as if virtual topics aren't being used a topic will be expected

yuanmwang-wf: oh yeah, good catch, will fix.

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	if len(data) > m.maxPublishSize {
+		return thrift.NewTTransportException(
+			TRANSPORT_EXCEPTION_REQUEST_TOO_LARGE,
+			fmt.Sprintf("Message exceeds %d bytes, was %d bytes", m.maxPublishSize, len(data)))
+	}
+
+	destination := m.formatStompPublishTopic(topic)
+	if err := m.conn.Send(destination, "application/octet-stream", data); err != nil {
```
brianshannan-wf: According to http://activemq.apache.org/stomp.html, we should add some headers
1. the persistent header. I think this should either be a flag to set or should default to true, I made this a flag in my java pull request.
2. the content-length header. With the stomp protocol, active mq apparently decides a message should be treated as binary if there's a content-length header, and should be treated as text otherwise. For interoperability with other languages, we want our messages treated as binary

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+
+// NewFStompSubscriberTransportFactory creates FStompSubscribeTransportFactory with the given stomp
+// connection and consumer name.
+func NewFStompSubscriberTransportFactory(conn *stomp.Conn, consumerPrefix string) *FStompSubscribeTransportFactory {
+	return &FStompSubscribeTransportFactory{conn: conn, consumerPrefix: consumerPrefix}
+}
+
+// NewFStompSubscriberTransportFactory creates FStompSubscribeTransportFactory with the given stomp
+// connection, consumer name and topic.
+func NewFStompSubscriberTransportFactoryWithQueue(conn *stomp.Conn, consumerPrefix string, queue string) *FStompSubscribeTransportFactory {
```
brianshannan-wf: nit: can we rename `queue` here to match the parameter in other function signatures?

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+
+// NewStompFSubscriberTransport creates a new FSubscriberTransport which is used for
+// pub/sub.
+func NewStompFSubscriberTransport(conn *stomp.Conn, consumerPrefix string) FSubscriberTransport {
+	return &fStompSubscriberTransport{conn: conn, consumerPrefix: consumerPrefix}
+}
+
+// NewStompFSubscriberTransport creates a new FSubscriberTransport which is used for
+// pub/sub with a topic.
+func NewStompFSubscriberTransportWithQueue(conn *stomp.Conn, consumerPrefix string, topic string) FSubscriberTransport {
```
brianshannan-wf: I don't think this function should exist, the topic should be passed into `Subscribe()`. If users look at this they might be confused why this constructor takes a topic as well as the subscribe call. Passing it into subscribe allows the transport to be reused.

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+	defer m.openMu.RUnlock()
+	return m.conn != nil && m.isSubscribed
+}
+
+// Unsubscribe unsubscribes from the destination.
+func (m *fStompSubscriberTransport) Unsubscribe() error {
+	m.openMu.Lock()
+	defer m.openMu.Unlock()
+	if !m.isSubscribed {
+		return nil
```
brianshannan-wf: This might be something good to add an info log for, as it's a case that really shouldn't happen so consumers might appreciate knowing about it

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+	m.isSubscribed = false
+	m.callback = nil
+	return nil
+}
+
+// Format subscription destination - subscribe to queue if consumer is using virtual topic, otherwise subscribe to a
+// regular topic.
+func (m *fStompSubscriberTransport) formatDestination(topic string) string {
+	subTopics := strings.SplitAfterN(m.consumerPrefix, ".", 3)
+	if len(subTopics) >= 3 && subTopics[0] == "Consumer." && subTopics[2] == "VirtualTopic." {
```
brianshannan-wf: IMO we should just pass a `useQueue` flag to prevent having to do this logic, as if we decide to change the virtual topic naming conventions in the future we'd have to change this. There's also other cases where people might want to use queues, and if we want to add queue support for publishers in the future, allowing a flag to be passed in would be more flexible

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+		case <-stopC:
+			return
+		case message, ok := <-m.sub.C:
+			if !ok {
+				logger().Errorf("frugal: error processing subscription messages, message channel closed unexpectedly")
+				return
+			}
+
+			if len(message.Body) < 4 {
+				continue
```
brianshannan-wf: This should have a log associated with it as it's an error condition, similar to https://github.com/Workiva/frugal/blob/master/lib/go/nats_scope_transport.go#L198

lib/go/stomp_transport_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"git.apache.org/thrift.git/lib/go/thrift"
+	"github.com/go-stomp/stomp"
+	"github.com/go-stomp/stomp/server"
+	"github.com/stretchr/testify/assert"
+)
+
+// Ensures stomp transport is able to open and close.
+func TestStompPublisherOpenPublish(t *testing.T) {
+	// starts a tcp server.
+	l, _ := net.Listen("tcp", ":0")
```
brianshannan-wf: What are the benefits to this approach over mocking out an interface?

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+func (m *fStompPublisherTransport) Open() error {
+	if m.conn == nil {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_NOT_OPEN, "frugal: mq transport not open")
+	}
+	return nil
+}
+
+// IsOpen returns true if the transport is open, false otherwise.
+func (m *fStompPublisherTransport) IsOpen() bool {
+	return m.conn != nil
```
tylerrinnan-wf: Is it sufficient to just ensure the stomp connection is not nil, or should we additionally check the status of connection here? https://github.com/go-stomp/stomp/blob/master/conn.go#L30

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+
+// Close closes the transport.
+func (m *fStompPublisherTransport) Close() error {
+	return nil
+}
+
+// GetPublishSizeLimit returns the maximum allowable size of a payload
+// to be published. A non-positive number is returned to indicate an
+// unbounded allowable size.
+func (m *fStompPublisherTransport) GetPublishSizeLimit() uint {
```
tylerrinnan-wf: > A non-positive number is returned to indicate an unbounded allowable size. 

How can a non-positive number be returned if the return type is a `uint` ?

brianshannan-wf: In this case zero is the only thing that satisfies that

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+	"strings"
+	"sync"
+
+	"git.apache.org/thrift.git/lib/go/thrift"
+	"github.com/go-stomp/stomp"
+)
+
+// FStompPublisherTransportFactory creates fStompPublisherTransports.
+type FStompPublisherTransportFactory struct {
+	conn           *stomp.Conn
```
tylerrinnan-wf: Wondering if this might be better as an interface so we're not directly tied to the `go-stomp` client since active-mq supports multiple protocols.

lib/go/stomp_transport.go
tylerrinnan-wf: Can we change the comment to be more explicit then and say `0` will be returned?

lib/go/stomp_transport.go
brianshannan-wf: I'm not sure how valuable this would be. I think the programatic apis for other protocols are different enough that we might as well make another transport at that point. The case I could see for an interface is for mocking in unit tests

lib/go/stomp_transport.go
yuanmwang-wf: Looks like [content-length header is present by default](https://github.com/go-stomp/stomp/blob/master/send_options.go#L19), so I've only added `persistent:true` here.

lib/go/stomp_transport.go
yuanmwang-wf: Removed.

lib/go/stomp_transport.go
yuanmwang-wf: Done in 4acde56

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,237 @@
...
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_ALREADY_OPEN, "frugal: stomp transport already has a subscription")
+	}
+
+	if topic == "" {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_UNKNOWN, "frugal: stomp transport cannot subscribe to empty topic")
+	}
+
+	var destination string
+	if m.useQueue {
+		destination = fmt.Sprintf("/queue/%s%s", m.consumerPrefix, topic)
```
brianshannan-wf: I think we need to format the frugal prefix as the middle piece.

brianshannan-wf:
I think we should just get this in after the last comment I made, I think everything else looks good and we'd benefit more from having this in the integration branch

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,252 @@
...
+func NewStompFSubscriberTransportWithQueue(conn *stomp.Conn, consumerPrefix string, topic string) FSubscriberTransport {
+	return &fStompSubscriberTransport{conn: conn, consumerPrefix: consumerPrefix, topic: topic}
+}
+
+// Subscribe sets the subscribe topic and opens the transport.
+func (m *fStompSubscriberTransport) Subscribe(topic string, callback FAsyncCallback) error {
+	m.openMu.Lock()
+	defer m.openMu.Unlock()
+
+	if m.conn == nil {
```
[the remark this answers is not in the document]
yuanmwang-wf: I think `isOpen` is in the publisher's interface, it's not available for subscribers.

lib/go/stomp_transport.go
yuanmwang-wf: Explicitly mentioning `0` now in all places.

brianshannan-wf:
+1
I think we should get this into the integration branch

lib/go/stomp_transport.go
yuanmwang-wf: Yeah I agree - I'm not sure if other protocols would be similar enough. About unit testing - I've looked around and it seems like generally it's recommended to use real connections if possible ([Example](https://speakerdeck.com/mitchellh/advanced-testing-with-go?slide=36))? 🤷‍♂️

tylerrinnan-wf: Makes sense, thanks.

tylerrinnan-wf:
+1, nice work!

tylerrinnan-wf:
Do we want to add an example using this publisher and subscriber or do that in another PR? @brianshannan-wf

brianshannan-wf:
I'm fine with another pr, I added some cross runner stuff in my java pr that this will need, so we can do it then

yuanmwang-wf:
Created [INFENG-6715: Add examples using go activemq transport in frugal](https://jira.atl.workiva.net/browse/INFENG-6715).

PR #1048, opened 2018-11-13
Java jms transport

brianshannan-wf:
Implements a JMS transport for frugal that could be used to communicate with active mq. I think all the major functionality exists, but might make some small changes. Some points
- Not super happy about how I ended up using the JMS session and producer/consumer in almost the same way (creating a session then immediately creating a producer, and also closing at the same time). The other possibility I see is sharing a session across multiple publishers/subscribers, but I was a little concerned as the documentation said sessions are not threadsafe, and I didn't want to coordinate that between transports. But I might have misunderstood how that API is supposed to work.
- The frugal transport API doesn't expose a good way to specify a message group per publish (for ordering), and I'm not sure how helpful providing one at transport construction time is going to be.
- Having a `topicPrefix` option makes using virtual topics in active mq much easier by specifying a topic prefix of `VirtualTopic.` for publishers and `Consumer.<me>.VirtualTopic` for subscribers.
- I tried to make this general enough so not to shoe horn ourselves into one thing.
- If I missed any major features you guys think would be useful, let me know.

@kevinsookocheff-wf @brettkail-wk @chadknight-wf if interested feedback is appreciated, enterprisey java isn't my strong suit

### DIFF
--- a/examples/java/src/main/java/examples/ActiveMQSubscriber.java
+++ b/examples/java/src/main/java/examples/ActiveMQSubscriber.java
@@ -1,0 +1,29 @@
+package examples;
+
+import com.workiva.frugal.protocol.FProtocolFactory;
+import com.workiva.frugal.provider.FScopeProvider;
+import com.workiva.frugal.transport.FJmsSubscriberTransport;
+import com.workiva.frugal.transport.FSubscriberTransportFactory;
+import org.apache.activemq.ActiveMQConnectionFactory;
+import org.apache.thrift.protocol.TBinaryProtocol;
+import v1.music.AlbumWinnersSubscriber;
+
+import javax.jms.Connection;
+
+public class ActiveMQSubscriber {
+    public static void main(String[] args) throws Exception {
+        FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
+
+        ActiveMQConnectionFactory connectionFactory = new ActiveMQConnectionFactory();
+        Connection connection = connectionFactory.createConnection();
+        connection.start();
+
+        FSubscriberTransportFactory subscriberFactory = new FJmsSubscriberTransport.Factory(connection, "Consumer.frugal-examples.VirtualTopic.");
+        FScopeProvider provider = new FScopeProvider(null, subscriberFactory, protocolFactory);
+
+        AlbumWinnersSubscriber.Iface subscriber = new AlbumWinnersSubscriber.Client(provider);
+        subscriber.subscribeWinner((ctx, album) -> System.out.println("You won! " + album));
+        subscriber.subscribeContestStart((ctx, albums) -> System.out.println("Contest started, available albums: " + albums));
+        System.out.println("Subscriber started...");
+    }
+}
--- a/examples/java/src/main/java/examples/ActiveMQSubscriber.java
+++ b/examples/java/src/main/java/examples/ActiveMQSubscriber.java
@@ -18,7 +18,7 @@ public class ActiveMQSubscriber {
         Connection connection = connectionFactory.createConnection();
         connection.start();
 
-        FSubscriberTransportFactory subscriberFactory = new FJmsSubscriberTransport.Factory(connection, "Consumer.frugal-examples.VirtualTopic.");
+        FSubscriberTransportFactory subscriberFactory = new FJmsSubscriberTransport.Factory(connection, "Consumer.frugal-examples.VirtualTopic.", true);
         FScopeProvider provider = new FScopeProvider(null, subscriberFactory, protocolFactory);
 
         AlbumWinnersSubscriber.Iface subscriber = new AlbumWinnersSubscriber.Client(provider);
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>LOCAL</version>
+    <version>2.23.0</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -55,6 +55,11 @@
             <version>4.5</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>activemq-client</artifactId>
+            <version>5.15.7</version>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
             <version>2.8.0</version>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -64,8 +64,9 @@
         </dependency>
         <dependency>
             <groupId>javax.jms</groupId>
-            <artifactId>jms</artifactId>
-            <version>1.1</version>
+            <artifactId>javax.jms-api</artifactId>
+            <version>2.0.1</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.google.code.gson</groupId>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>LOCAL</version>
+    <version>2.23.0</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
@@ -38,6 +38,14 @@
         <url>http://github.com/Workiva/frugal</url>
     </scm>
 
+    <repositories>
+        <repository>
+            <id>repository.jboss.org-public</id>
+            <name>JBoss.org Maven repository</name>
+            <url>https://repository.jboss.org/nexus/content/groups/public</url>
+        </repository>
+    </repositories>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.thrift</groupId>
@@ -55,9 +63,9 @@
             <version>4.5</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.activemq</groupId>
-            <artifactId>activemq-client</artifactId>
-            <version>5.15.7</version>
+            <groupId>javax.jms</groupId>
+            <artifactId>jms</artifactId>
+            <version>1.1</version>
         </dependency>
         <dependency>
             <groupId>com.google.code.gson</groupId>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>2.23.0</version>
+    <version>LOCAL</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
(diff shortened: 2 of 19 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/java/pom.xml
```diff
@@ -54,6 +62,11 @@
             <artifactId>httpclient</artifactId>
             <version>4.5</version>
         </dependency>
+        <dependency>
+            <groupId>javax.jms</groupId>
+            <artifactId>jms</artifactId>
+            <version>1.1</version>
```
brettkail-wk: It's somewhat unfortunate to add an unconditional dependency, even if only to an API JAR.  If we don't want to create a separate `frugal-jms` artifact(?), then it seems like it would be better to use `<scope>provided</scope>` and then require anyone that wants JMS support to pull in the JMS API JAR themselves.

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public synchronized void open() throws TTransportException {
+        // TODO test
+        if (isOpen()) {
+            LOGGER.info("jms transport already open, returning");
+            return;
+        }
+
+        try {
+            session = connection.createSession(false, Session.CLIENT_ACKNOWLEDGE);
```
brettkail-wk: I also don't know JMS well enough, but `Session` is at least documented as a lightweight object, so it seems to be that a new session should be created on a per-publish/subscribe basis.

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,145 @@
...
+                    LOGGER.error("unhandled message type '%s'", message.getClass().getName());
+                    return;
+                }
+
+                if (payload.length < 4) {
+                    LOGGER.warn("discarding invalid message frame, length less than four");
+                    return;
+                }
+                try {
+                    // TODO better way than copying? what we do for other transports
```
brettkail-wk: Use `new TMemoryInputTransport(payload, 4, payload.length - 4)`?

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public synchronized void close() {
+        if (!isOpen()) {
+            LOGGER.info("jms transport already closed, returning");
+            return;
+        }
+        // TODO test
+        try {
+            producer.close();
+            session.close();
```
brettkail-wk: try-with-resource ensures the second is closed even if closing the first one fails:
```
try (MessageProducer closeProducer = producer) {
  session.close();
}
```
(Later Java supports `try (producer) { session.close(); }` without the dummy var.)

lib/java/pom.xml
brianshannan-wf: That's definitely doable

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
brianshannan-wf: My interpretation of the documentation was a little different, but I might be wrong. From the javadoc for JMS I found 
```
Although it may allocate provider resources outside the Java virtual machine (JVM), it is considered a lightweight JMS object.
```
And was trying to figure out was that meant, and some [activemq](http://activemq.apache.org/how-do-i-use-jms-efficiently.html) documentation had the following
```
The reason is that each create & destroy of the above objects typically requires an individual request & response with the JMS broker to ensure it worked. e.g. creating a connection, session, producer, then sending a message, then closing everything down again - could result in 7 request-responses with the server
```
Which might be specific to activemq, as I couldn't find a general statement about it.

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
brianshannan-wf: Yeah that's definitely better, not sure how I missed that. Thanks!

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
brettkail-wk: My best guess is that either `createSession` is supposed to lightweight, or it can be made lightweight through pooling.  It looks like ActiveMQ has `PooledConnection` builtin, and it looks like it pools sessions, so I would be inclined to try that and avoid long-lived `Session` like this.

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,142 @@
...
+                    LOGGER.warn("discarding invalid message frame, length less than four");
+                    return;
+                }
+                try {
+                    callback.onMessage(
+                            new TMemoryInputTransport(payload, 4, payload.length - 4)
+                    );
+                } catch (TException e) {
+                    LOGGER.error("error executing user provided callback", e);
+                }
```
chadknight-wf: Should we also return early in this error case, before acknowledging the message? If a handler throws a TException, I think we expect that the message is not acked.

brianshannan-wf: Yes, good catch

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,143 @@
...
+                    LOGGER.warn("discarding invalid message frame, length less than four");
+                    return;
+                }
+                try {
+                    callback.onMessage(
+                            new TMemoryInputTransport(payload, 4, payload.length - 4)
+                    );
+                } catch (TException e) {
+                    LOGGER.error("error executing user provided callback", e);
+                    return;
```
brettkail-wk: I don't know much about queueing, and I understand that unconditionally acknowledging on failure is wrong, but unconditionally returning here also seems wrong.  If there is unconditional error in the handler (e.g., message sent in BINARY vs COMPAT format, handler always throws an NPE for whatever reason), then this will cause the message to eventually reach visibility timeout and retry (indefinitely?).  Is there a way to mark the message as failing to handle, like resetting its retry interval, moving to a dead letter queue, etc.?  Even then, it seems like you somehow want to define exceptions that allow callers to explicitly control the behavior (e.g., if they know it's a retryable error, then they can signal to do that without logging a noisy error and instead of moving to DLQ).

brianshannan-wf: Here's my general thoughts on this, which might be wrong.
- Redelivery can usually be set on the connection, for example activemq allows configuring a [redelivery policy](http://activemq.apache.org/redelivery-policy.html) to configure settings like how many times should a message be redelivered, the frequency, etc.
- If it's unclear whether the error is retryable at this level in the code, we should err on the side of not acking the message. Hence in general if an exception is returned, we don't ack the message.
- If the user provided callback identifies an error isn't retryable, it should do something to handle that error, and would then not throw that exception, causing frugal to ack the message. Not throwing an exception in this case seems easier than defining/throwing a specific exception, but does lead to a simplistic control flow. I have not found a way in jms to ack a message in such a way that tells the broker that the message was received, but should be sent to the dead letter queue immediately. I might have missed it in the api though.

This should all probably be documented better as well.

brettkail-wk: That makes sense to me, thanks.  Better documentation is also welcome :-).  I also don't know enough about JMS to know whether it's possible to force redelivery.

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
blakeroberts-wk: What problems do you foresee with long-lived sessions? Would these issues be resolved using a connection pool?

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,143 @@
...
+    MessageProducer producer;
+
+    FJmsPublisherTransport(Connection connection, String topicPrefix, boolean durablePublishes) {
+        this.connection = connection;
+        this.topicPrefix = topicPrefix;
+        this.durablePublishes = durablePublishes;
+    }
+
+    /**
+     * An FPublisherTransportFacory implementation which creates
```
tylerrinnan-wf: Small nit:
```suggestion
     * An FPublisherTransportFactory implementation which creates
```

examples/java/src/main/java/examples/ActiveMQSubscriber.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ActiveMQSubscriber {
+    public static void main(String[] args) throws Exception {
+        FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
+
+        ActiveMQConnectionFactory connectionFactory = new ActiveMQConnectionFactory();
+        Connection connection = connectionFactory.createConnection();
+        connection.start();
+
+        FSubscriberTransportFactory subscriberFactory = new FJmsSubscriberTransport.Factory(connection, "Consumer.frugal-examples.VirtualTopic.");
```
tylerrinnan-wf:
```suggestion
        FSubscriberTransportFactory subscriberFactory = new FJmsSubscriberTransport.Factory(connection, "Consumer.frugal-examples.VirtualTopic.", true);
```

tylerrinnan-wf:
Ran the example with a local activemq running. 

+1

scripts/skynet/cross/setup_java.sh
```diff
@@ -9,7 +9,8 @@ if [ -z "${IN_SKYNET_CLI+yes}" ]; then
 else
     cd ${FRUGAL_HOME}/lib/java
     mvn clean verify -q
-    mv target/frugal-2.20.0.jar ${FRUGAL_HOME}/test/integration/java/frugal-integration-test/frugal.jar
+    ls target
```
yuanmwang-wf: nit: is this intentional?

scripts/skynet/skynet_setup.sh
```diff
@@ -16,3 +16,10 @@ cd $GOPATH/src/github.com/Workiva/frugal && go install
 
 # Start gnatsd
 gnatsd &
+
+# TODO this install should be in messaging-docker-images
+wget https://archive.apache.org/dist/activemq/5.15.2/apache-activemq-5.15.2-bin.tar.gz
```
yuanmwang-wf: nit: is it possible to download 5.15.6 so it [matches the broker version](https://github.com/Workiva/messaging-mq-broker/blob/master/cloudformation/messaging-mq-broker.yaml#L100) we're using?

test/integration/crossrunner/load.go
```diff
@@ -18,6 +18,7 @@ import (
 	"io/ioutil"
 	"os"
 	"time"
+	"fmt"
```
yuanmwang-wf: nit: is this sorted correctly?

yuanmwang-wf:
+1

PR #1052, opened 2018-11-26
INFENG-6670 java: catch throwable rather than TException in nats subscriber

brianshannan-wf:
I think this will fix the problem of RuntimeExceptions not being logged.

@Workiva/messaging-pp @Workiva/product2 @brettkail-wk

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
@@ -122,8 +122,8 @@ public class FNatsSubscriberTransport implements FSubscriberTransport {
                 callback.onMessage(
                         new TMemoryInputTransport(Arrays.copyOfRange(msg.getData(), 4, msg.getData().length))
                 );
-            } catch (TException e) {
-                LOGGER.error("error executing user provided callback", e);
+            } catch (Throwable t) {
+                LOGGER.error("error executing user provided callback", t);
             }
         });
     }


aviary3-wk:
## Security Insights
<details>
<summary>(1) Image validation issues were detected</summary>
</details>

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -122,8 +122,8 @@ public void subscribe(String topic, FAsyncCallback callback) throws TException {
                 callback.onMessage(
                         new TMemoryInputTransport(Arrays.copyOfRange(msg.getData(), 4, msg.getData().length))
                 );
-            } catch (TException e) {
-                LOGGER.error("error executing user provided callback", e);
+            } catch (Throwable t) {
+                LOGGER.error("error executing user provided callback", t);
```
brettkail-wk: No test that `RuntimeException` and `Error` don't escape?  I would probably capitalize also.

brettkail-wk: I guess you could, but I think you could just add `Handler.runtimeException` and `Handler.error` fields, and then extend the existing `testSubscribe` to set those fields instead of `exception`.

brianshannan-wf: Oh right, that's definitely easier

yuanmwang-wf:
+1

PR #1055, opened 2018-11-27
INFENG-6369: python asyncio activemq frugal transport

yuanmwang-wf:
**Overview**
Similar to how activemq transport is implemented in go and java, the transport uses connections created by [stomp.py](https://github.com/jasonrbriggs/stomp.py) to communicate with the broker.

**Reviewers**
@Workiva/product2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,6 @@ ARG BUILD_ARTIFACTS_ARTIFACTORY=/go/src/github.com/Workiva/frugal/frugal-*.jar
 ARG BUILD_ARTIFACTS_PUB=/go/src/github.com/Workiva/frugal/frugal.pub.tgz
 ARG BUILD_ARTIFACTS_TEST_RESULTS=/go/src/github.com/Workiva/frugal/test_results/*
 
-FROM scratch
+FROM alpine:latest
 COPY --from=build /go/src/github.com/Workiva/frugal/frugal /bin/frugal
 ENTRYPOINT ["frugal"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,6 @@ ARG BUILD_ARTIFACTS_ARTIFACTORY=/go/src/github.com/Workiva/frugal/frugal-*.jar
 ARG BUILD_ARTIFACTS_PUB=/go/src/github.com/Workiva/frugal/frugal.pub.tgz
 ARG BUILD_ARTIFACTS_TEST_RESULTS=/go/src/github.com/Workiva/frugal/test_results/*
 
-FROM alpine:latest
+FROM scratch
 COPY --from=build /go/src/github.com/Workiva/frugal/frugal /bin/frugal
 ENTRYPOINT ["frugal"]
--- a/lib/python/frugal/aio/transport/__init__.py
+++ b/lib/python/frugal/aio/transport/__init__.py
@@ -17,7 +17,6 @@ from .nats_scope_transport import FNatsSubscriberTransportFactory
 from .nats_scope_transport import FNatsSubscriberTransport
 from .nats_transport import FNatsTransport
 from .http_transport import FHttpTransport
-from .stomp_transport import FStompConnectionListener
 from .stomp_transport import FStompPublisherTransportFactory
 from .stomp_transport import FStompPublisherTransport
 from .stomp_transport import FStompSubscriberTransportFactory
@@ -33,7 +32,6 @@ __all__ = [
     'FNatsPublisherTransport',
     'FNatsSubscriberTransportFactory',
     'FNatsSubscriberTransport',
-    'FStompConnectionListener',
     'FStompPublisherTransportFactory',
     'FStompPublisherTransport',
     'FStompSubscriberTransportFactory',
--- a/lib/python/frugal/aio/transport/__init__.py
+++ b/lib/python/frugal/aio/transport/__init__.py
@@ -17,17 +17,25 @@ from .nats_scope_transport import FNatsSubscriberTransportFactory
 from .nats_scope_transport import FNatsSubscriberTransport
 from .nats_transport import FNatsTransport
 from .http_transport import FHttpTransport
+from .stomp_transport import FStompConnectionListener
+from .stomp_transport import FStompPublisherTransportFactory
+from .stomp_transport import FStompPublisherTransport
+from .stomp_transport import FStompSubscriberTransportFactory
+from .stomp_transport import FStompSubscriberTransport
 
 
 __all__ = [
     'FTransportBase',
     'FAsyncTransport',
     'FNatsTransport',
-    'FNatsScopeTransportFactory',
-    'FNatsScopeTransport',
     'FHttpTransport',
     'FNatsPublisherTransportFactory',
     'FNatsPublisherTransport',
     'FNatsSubscriberTransportFactory',
     'FNatsSubscriberTransport',
+    'FStompConnectionListener',
+    'FStompPublisherTransportFactory',
+    'FStompPublisherTransport',
+    'FStompSubscriberTransportFactory',
+    'FStompSubscriberTransport'
 ]
--- a/lib/python/frugal/aio/transport/stomp_transport.py
+++ b/lib/python/frugal/aio/transport/stomp_transport.py
@@ -1,0 +1,234 @@
+# Copyright 2017 Workiva
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import inspect
+from uuid import uuid4
+
+from stomp import Connection
+from stomp import ConnectionListener
+from thrift.transport.TTransport import TMemoryBuffer
+from thrift.transport.TTransport import TTransportException
+
+from frugal.exceptions import TTransportExceptionType
+from frugal.transport import FPublisherTransport
+from frugal.transport import FPublisherTransportFactory
+from frugal.transport import FSubscriberTransport
+from frugal.transport import FSubscriberTransportFactory
+
+
+FRUGAL_PREFIX = 'frugal.'
+
+
+class FStompPublisherTransportFactory(FPublisherTransportFactory):
+    """
+    FStompPublisherTransportFactory is used to create
+    FStompPublisherTransports.
+    """
+
+    def __init__(self, stomp_conn: Connection, topic_prefix: str,
+                 max_message_size: int = 32 * 1024 * 1024):
+        self._stomp_conn = stomp_conn
+        self._topic_prefix = topic_prefix
+        self._max_message_size = max_message_size
+
+    def get_transport(self) -> FPublisherTransport:
+        """
+        Get a new FStompPublisherTransport.
+        """
+        return FStompPublisherTransport(
+            self._stomp_conn, self._topic_prefix, self._max_message_size)
+
+
+class FStompPublisherTransport(FPublisherTransport):
+    """
+    FStompPublisherTransport is used exclusively for pub/sub scopes.
+    Publishers use it to publish to a topic. Messaging brokers that support
+    stomp protocol can be used as the underlying bus.
+    """
+
+    def __init__(self, stomp_conn: Connection, topic_prefix: str,
+                 max_message_size: int):
+        super().__init__(max_message_size)
+        self._stomp_conn = stomp_conn
+        self._topic_prefix = topic_prefix
+
+    async def open(self):
+        """
+        Open the stomp publisher in preparation for publishing.
+        """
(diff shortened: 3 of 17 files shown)


aviary2-wf:
## Security Insights
<details>
<summary>(1) Image validation issues were detected</summary>
</details>

## Action Items
---

lib/python/frugal/aio/transport/__init__.py
```diff
@@ -17,17 +17,25 @@
...
+from .stomp_transport import FStompSubscriberTransportFactory
+from .stomp_transport import FStompSubscriberTransport
 
 
 __all__ = [
     'FTransportBase',
     'FAsyncTransport',
     'FNatsTransport',
-    'FNatsScopeTransportFactory',
-    'FNatsScopeTransport',
```
yuanmwang-wf: I don't think these two are currently importable anyway so I removed them.

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,234 @@
...
+    async def close(self):
+        """
+        Close the stomp publisher transport and disconnect from the stomp
+        broker.
+        :return:
+        """
+        if not self.is_open():
+            return
+
+        await self._stomp_conn.disconnect()
```
yuanmwang-wf: I am not sure if `close` should disconnect from the broker since it does not connect in `open` - I think the actual connect should be performed by messaging sdk and then the established connection should be passed to frugal. Should frugal then be responsible for disconnecting the connection or should it be managed outside of frugal?

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,234 @@
...
+            )
+        destination = '/topic/{topic_prefix}{frugal_prefix}{topic}'.format(
+            topic_prefix=self._topic_prefix,
+            frugal_prefix=FRUGAL_PREFIX,
+            topic=topic
+        )
+        await self._stomp_conn.send(destination,
+                                    data,
+                                    content_type='application/octet-stream',
+                                    headers={'persistent': 'true'})
```
yuanmwang-wf: Same with go-stomp, content-length header is added automatically by [stomp.py](https://github.com/jasonrbriggs/stomp.py/blob/b04c754ec5e9401323e30fdd495bb3cc1731383d/stomp/protocol.py#L217)

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,234 @@
...
+                TTransportExceptionType.ALREADY_OPEN,
+                'stomp connection already subscribed to topic')
+
+        if not topic:
+            raise TTransportException(
+                TTransportExceptionType.UNKNOWN,
+                'stomp transport cannot subscribe to empty topic')
+
+        msg_listener = FStompConnectionListener(callback, self._stomp_conn)
+        sub_id = str(uuid4())
```
yuanmwang-wf: Looks like the connection needs to provide a subscription id when subscribing, I'm not sure if frugal should generate a uuid like this or should this be a parameter so it can be customizable in messaging sdk.

lib/python/frugal/transport/scope_transport.py
```diff
@@ -54,7 +54,7 @@ def publish(self, topic, data):
 
     def _check_publish_size(self, data):
         """
-        Returns True if the data is of permissible size, False otherwise.
+        Returns True if the length of data is over the max message size.
```
yuanmwang-wf: I think the current dosctring is wrong so I fixed it here.

lib/python/frugal/aio/transport/__init__.py
brianshannan-wf: Oof yeah, I think those are from 1.x

lib/python/frugal/transport/scope_transport.py
yuanmwang-wf: It should at least be `Returns False if the data is of permissible size` though? 

For example here it's raising an exception if this check returns True. https://github.com/Workiva/frugal/blob/master/lib/python/frugal/aio/transport/nats_scope_transport.py#L88

yuanmwang-wf:
>Unfortunately we can't use this stomp library for our asyncio library as it uses the standard threading library and not asyncio coroutines.

Ahh you're right... This is the only asyncio stomp library I could find: https://github.com/pedrokiefer/aiostomp. Can you take a look and see if this will work here? @brianshannan-wf

lib/python/frugal/transport/scope_transport.py
brianshannan-wf: Okay, I know what's wrong. This function was intended to be `self._max_message_size > len(data) > 0`, as right now the minimum size isn't checked. I think the best course of action here is to update the doc comment (as you did) and fix the logic in the function so usages of this function won't break

yuanmwang-wf: Sounds good 👍

brianshannan-wf:
As long as it has all the features we need and uses asyncio coroutines it should be fine. It doesn't have a ton of stars, but it might be the best available thing

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class FStompSubscriberTransportFactory(FSubscriberTransportFactory):
+    """
+    FStompSubscriberTransportFactory is used to create
+    FStompSubscriberTransports.
+    """
+
+    def __init__(self, stomp_conn: Connection, consumer_prefix: str,
+                 use_queue: bool):
```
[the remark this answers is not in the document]
yuanmwang-wf: I've added a default value for `use_queue`, I can't think of a sane default value for `consumer_prefix` though since it would be like `Consumer.A.` and I don't think there's a good default value for the 'A' part.

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,234 @@
...
+        self._stomp_conn = stomp_conn
+        self._consumer_prefix = consumer_prefix
+        self._use_queue = use_queue
+        self._sub_id = None
+
+    @property
+    def _is_subscribed(self):
```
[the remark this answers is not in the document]
yuanmwang-wf: Agreed, this shouldn't be necessary, updated in 1cbf7ab

yuanmwang-wf:
I've changed the underlying stomp library to [AioStomp](https://github.com/pedrokiefer/aiostomp) - I can't find a better asyncio stomp library, so far it seems to offer what frugal needs.

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,190 @@
...
+                TTransportExceptionType.ALREADY_OPEN,
+                'stomp connection already subscribed to topic')
+
+        if not topic:
+            raise TTransportException(
+                TTransportExceptionType.UNKNOWN,
+                'stomp transport cannot subscribe to empty topic')
+
+        async def msg_handler(frame, _):
+            ret = callback(TMemoryBuffer(frame.body[4:]))
```
yuanmwang-wf: Looks like [AioStomp handles ack messages if subscription is in `client-individual` mode](https://github.com/pedrokiefer/aiostomp/blob/e455bf3f65ddbb9080cb7d03ae23789c20eea93b/aiostomp/aiostomp.py#L365) as long as callback returns a non-falsy value. I'm not sure if this will work here. Does frugal require consumers' callback functions to return non-falsy values? If not frugal might need to wrap the callback here and return True if the original callback did not raise an exception?

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,190 @@
...
+
+class FStompPublisherTransport(FPublisherTransport):
+    """
+    FStompPublisherTransport is used exclusively for pub/sub scopes.
+    Publishers use it to publish to a topic. Messaging brokers that support
+    stomp protocol can be used as the underlying bus.
+    """
+
+    def __init__(self, stomp_client: AioStomp,
+                 topic_prefix: str='VirtualTopic.',
```
brianshannan-wf: I think what makes the most sense to default this to an empty string here, for a couple reasons.
1. We can default to virtual topics in the messaging-sdk if we want to default to that
2. There's not a good default for subscriber transports, leading to a mismatch of what's set
3. Not all stomp brokers support virtual topics, if we ever change what broker we want to use

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+
+class FStompSubscriberTransportFactory(FSubscriberTransportFactory):
+    """
+    FStompSubscriberTransportFactory is used to create
+    FStompSubscriberTransports.
+    """
+
+    def __init__(self, stomp_client: AioStomp, consumer_prefix: str,
```
brianshannan-wf: Same comment about prefix as above

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,190 @@
...
+                 use_queue: bool=True):
```
brianshannan-wf: Imo this should default to `False` for many of the same reasons the prefix should be options

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,190 @@
...
+            raise TTransportException(
+                type=TTransportExceptionType.REQUEST_TOO_LARGE,
+                message='Message exceeds max message size'
+            )
+        destination = '/topic/{topic_prefix}{frugal_prefix}{topic}'.format(
+            topic_prefix=self._topic_prefix,
+            frugal_prefix=FRUGAL_PREFIX,
+            topic=topic
+        )
+        self._stomp_client.send(
```
brianshannan-wf: It seems a little weird to me that a send operation on an asyncio library wouldn't be async. I spent some time looking through the library and it's... unclear. Would be good to get either (or both) examples or skynet tests written for this to verify it works, given I couldn't quickly ascertain whether this returns a future

lib/python/frugal/aio/transport/stomp_transport.py
brianshannan-wf: Frugal doesn't do anything with the return value, wrapping the callback seems like a good solution, as we're already doing that here to provide the correct input type.

brianshannan-wf:
Overall looks good, I'd like to have examples/skynet tests similar to the java pr so it's easier to qa this.

yuanmwang-wf:
Confirmed examples now work with local activemq broker, also confirmed `send` is indeed not async 
:(

Working on skynet tests now.

yuanmwang-wf:
@Workiva/product2 I've added examples and skynet tests for both publisher and subscriber with activemq, should be ready for review again.

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class FStompPublisherTransportFactory(FPublisherTransportFactory):
+    """
+    FStompPublisherTransportFactory is used to create
+    FStompPublisherTransports.
+    """
+
+    def __init__(self, stomp_conn: Connection, topic_prefix: str,
+                 max_message_size: int = 32 * 1024 * 1024):
```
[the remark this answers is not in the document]
blakeroberts-wk: @brianshannan-wf https://github.com/PyCQA/pycodestyle/blob/bd17ca651a57f8cdd278a8fac84d6a09a178a3eb/pycodestyle.py#L928
👀

Dockerfile
```diff
@@ -35,6 +35,6 @@ ARG BUILD_ARTIFACTS_ARTIFACTORY=/go/src/github.com/Workiva/frugal/frugal-*.jar
 ARG BUILD_ARTIFACTS_PUB=/go/src/github.com/Workiva/frugal/frugal.pub.tgz
 ARG BUILD_ARTIFACTS_TEST_RESULTS=/go/src/github.com/Workiva/frugal/test_results/*
 
-FROM scratch
+FROM alpine:latest
```
brianshannan-wf: What was this changed?

yuanmwang-wf:
@brianshannan-wf the base image change was recommended by buildtools because there was a bug in workiva build - it would not send a notification to skynet if the base image is scratch so skynet tests wouldn't automatically start.

brianshannan-wf:
I think @travisreed-wf fixed that bug

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,192 @@
...
+        No-op. Client close should be handled outside of frugal.
+        :return:
+        """
+        pass
+
+    def is_open(self) -> bool:
+        """
+        No-op.
+        """
+        return True
```
danielnovotny-wf: Is this a typical pattern for other transports to default this to `True`?

yuanmwang-wf: I thinks so:

https://github.com/Workiva/frugal/blob/master/lib/python/frugal/aio/transport/http_transport.py#L58

https://github.com/Workiva/frugal/blob/master/lib/go/http_transport.go#L225

yuanmwang-wf:
Looks like `FROM scratch` works now 🎉

danielnovotny-wf:
+1

brianshannan-wf:
+1

PR #1067, opened 2018-12-21
ActiveMQ transports for go, java and python asyncio 

yuanmwang-wf:
### Story:
Final PR to merge all stomp transports to develop branch

### How To Test:
Workiva build and skynet should pass. 

Examples in all languages should work. 

Public APIs across all languages should be consistent. 

#### Reviewers:
@Workiva/product2


examples/java/src/main/java/examples/ActiveMQPublisher.java
```diff
@@ -0,0 +1,62 @@
+package examples;
+
+import com.sun.istack.internal.logging.Logger;
```
tylerrinnan-wf: I wasn't able to get `mvn clean install` to work until I removed this line and line 24. The logger isn't being used so can we just nix these lines?

examples/go/stompSubscriber/stompSubsriber.go
```diff
@@ -0,0 +1,46 @@
+package main
```
tylerrinnan-wf: This file was named "subsriber" on accident.

tylerrinnan-wf:
Running the examples locally, I'm able to get all the languages to work with themselves, but wasn't able to get say a Java stompPublisher to work with a Golang stompSubscriber.

yuanmwang-wf:
I used my fork because build was failing & there were some merge conflicts, I'll close this PR and use workiva/amazon-mq as base branch instead.

PR #1071, opened 2019-01-02
ActiveMQ transports for go, java and python asyncio

yuanmwang-wf:
### Story:
PR to merge all activemq transports from integration branch to develop. 

#### Reviewers:
@Workiva/product2


aviary2-wf:
## Security Insights
<details>
<summary>(2) Image validation issues were detected</summary>
<li>Application appears to run using root user.</li>
</details>

## Action Items
* [Specify a non-root USER](https://wiki.atl.workiva.net/display/SECURITY/Security+Requirements+for+Building+Containers#SecurityRequirementsforBuildingContainers-RunningwithNon-rootUser) in your deploy stage.
---

brianshannan-wf:
The examples don't work cross language correctly because of different virtual topic settings, I'll get that fixed

examples/go/stompSubscriber/stompSubsriber.go
```diff
@@ -0,0 +1,46 @@
+package main
```
brianshannan-wf: Typo in the file name

examples/java/pom.xml
```diff
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>com.workiva</groupId>
             <artifactId>frugal</artifactId>
-            <version>2.25.2</version>
+            <version>LOCAL</version>
```
brianshannan-wf: Change this

examples/java/src/main/java/examples/ActiveMQPublisher.java
```diff
@@ -0,0 +1,62 @@
+package examples;
+
+import com.sun.istack.internal.logging.Logger;
```
brianshannan-wf: Wrong logging library, should be `org.slf4j.Logger` I think

examples/java/src/main/java/examples/ActiveMQPublisher.java
```diff
@@ -0,0 +1,62 @@
...
+    public static void main(String[] args) throws Exception {
+        Logger.getLogger(FJmsPublisherTransport.class).setLevel(Level.ALL);
+
+        FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
+
+        ActiveMQConnectionFactory connectionFactory = new ActiveMQConnectionFactory();
+        Connection connection = connectionFactory.createConnection();
+        connection.start();
+
+        FPublisherTransportFactory publisherFactory = new FJmsPublisherTransport.Factory(connection, "VirtualTopic.", true);
```
brianshannan-wf: This shouldn't use virtual topics to match other languages

examples/java/src/main/java/examples/ActiveMQSubscriber.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ActiveMQSubscriber {
+    public static void main(String[] args) throws Exception {
+        FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
+
+        ActiveMQConnectionFactory connectionFactory = new ActiveMQConnectionFactory();
+        Connection connection = connectionFactory.createConnection();
+        connection.start();
+
+        FSubscriberTransportFactory subscriberFactory = new FJmsSubscriberTransport.Factory(connection, "Consumer.frugal-examples.VirtualTopic.", true);
```
brianshannan-wf: Shouldn't use virtual topics to match other examples

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,143 @@
...
+        } finally {
+            producer = null;
+            session = null;
+        }
+    }
+
+    @Override
+    public int getPublishSizeLimit() {
+        // TODO check this number
+        return 32 * 1024 * 1024;
```
yuanmwang-wf: not sure if this is a big deal - I think publish size limits are configurable (consumer has to set this value instead of being hardcoded), should this be consistent between all languages?

examples/go/stompSubscriber/stompSubsriber.go
tylerrinnan-wf: rename this file to `stompSubscriber`

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,237 @@
...
+// FStompPublisherTransportFactory creates fStompPublisherTransports.
+type FStompPublisherTransportFactory struct {
+	conn           *stomp.Conn
+	maxPublishSize int
+	topicPrefix    string
+}
+
+// NewFStompPublisherTransportFactory creates an FStompPublisherTransportFactory using the
+// provided stomp connection.
+func NewFStompPublisherTransportFactory(conn *stomp.Conn, maxPublishSize int, topicPrefix string) *FStompPublisherTransportFactory {
```
brianshannan-wf: Should we make this a factory? We do that for frugal java, it's a little more burdensome to use but more extensible for the future

examples/java/src/main/java/examples/ActiveMQPublisher.java
tylerrinnan-wf: Remove this and line 24 so that maven can clean install without error.

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,237 @@
...
+// fStompPublisherTransport implements FPublisherTransport.
+type fStompPublisherTransport struct {
+	conn           *stomp.Conn
+	maxPublishSize int
+	topicPrefix    string
+}
+
+// NewStompFPublisherTransport creates a new FPublisherTransport which is used for
+// publishing using stomp protocol with scopes.
+func NewStompFPublisherTransport(conn *stomp.Conn, maxPublishSize int, topicPrefix string) FPublisherTransport {
```
brianshannan-wf: Does this need to be public? I'm thinking it doesn't, but not 100% sure

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,237 @@
...
+	return &fStompPublisherTransport{conn: conn, maxPublishSize: maxPublishSize, topicPrefix: topicPrefix}
+}
+
+// Open initializes the transport.
+func (m *fStompPublisherTransport) Open() error {
+	if m.conn == nil {
+		return thrift.NewTTransportException(TRANSPORT_EXCEPTION_NOT_OPEN, "frugal: mq transport not open")
```
brianshannan-wf: s/mq/stomp

lib/go/stomp_transport.go
yuanmwang-wf: I can see it either way, nats in go also uses a factory too though, so maybe keep so it stays consistent in the same language as well?

brianshannan-wf: Err I meant a builder for the factory

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,143 @@
...
+     * An FPublisherTransportFactory implementation which creates
+     * FPublisherTransports backed by a JMS connection.
+     */
+    public static class Factory implements FPublisherTransportFactory {
+
+        private final Connection connection;
+        private final String topicPrefix;
+        private final boolean durablePublishes;
+
+        // TODO should we make a builder for this?
```
brianshannan-wf: See builder discussion for go

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Override
+    public synchronized boolean isOpen() {
+        return session != null && producer != null;
+    }
+
+    @Override
+    public synchronized void open() throws TTransportException {
+        // TODO test
```
brianshannan-wf: remove

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
brianshannan-wf: I think configurable makes sense

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public int getPublishSizeLimit() {
+        // TODO check this number
+        return 32 * 1024 * 1024;
+    }
+
+    @Override
+    public synchronized void publish(String topic, byte[] payload) throws TTransportException {
+        // TODO test
```
brianshannan-wf: Remove this

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+import static com.workiva.frugal.transport.FNatsTransport.FRUGAL_PREFIX;
+
+/**
+ * FJmsSubscriberTransport implements FSubscriberTransport by utilizing a JSM
+ * connection.
+ */
+public class FJmsSubscriberTransport implements FSubscriberTransport {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FJmsSubscriberTransport.class);
+
+    // TODO sessions aren't threadsafe, do we need to do more than sync on sub/unsub?
```
brianshannan-wf: I think synchronizing on sub/unsub is all that's needed, but should check

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * An FSubscriberTransportFacory implementation which creates
+     * FSubscriberTransports backed by a JMS connection.
+     */
+    public static class Factory implements FSubscriberTransportFactory {
+        private final Connection connection;
+        private final String topicPrefix;
+        private final boolean useQueues;
+
+        // TODO should we make a builder for this?
```
brianshannan-wf: See builder discussion in go

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Override
+    public synchronized boolean isSubscribed() {
+        return session != null && consumer != null;
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * If an exception is raised by the provided callback, the message will
```
brianshannan-wf: We should have this documentation in every language, it's good information on how error handling is done in subscriber callbacks

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+     * not be acked with the broker. This behaviour allows the message to be
+     * redelivered and processing to be attempted again. If an exception is
+     * not raised by the provided callback, the message will be acked. This is
+     * used if processing succeeded, or if it's apparent processing will never
+     * succeed, as the message won't continue to be redelivered.
+     */
+    @Override
+    public synchronized void subscribe(String topic, FAsyncCallback callback) throws TException {
+        // TODO test
```
brianshannan-wf: Remove this

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+            throw new TTransportException(TTransportException.ALREADY_OPEN, "jms client already subscribed");
+        }
+
+        if (topic == null || "".equals(topic)) {
+            throw new TTransportException("subscribe subject cannot be empty");
+        }
+        String formattedTopic = getFormattedTopic(topic);
+
+        try {
+            // TODO I think these are the defaults we want
```
brianshannan-wf: Double check then remove this

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+                destination = session.createQueue(formattedTopic);
+            } else {
+                destination = session.createTopic(formattedTopic);
+            }
+
+            consumer = session.createConsumer(destination);
+            consumer.setMessageListener(message -> {
+                LOGGER.debug("received a message on topic '%s'", formattedTopic);
+
+                // TODO do we need to handle other message types?
```
brianshannan-wf: I don't think we do, remove this

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
```diff
@@ -0,0 +1,164 @@
...
+            Destination destination;
+            if (useQueues) {
+                destination = session.createQueue(formattedTopic);
+            } else {
+                destination = session.createTopic(formattedTopic);
+            }
+
+            consumer = session.createConsumer(destination);
+            consumer.setMessageListener(message -> {
+                LOGGER.debug("received a message on topic '%s'", formattedTopic);
```
brianshannan-wf: Should discuss if it's worth adding simple debug logs like this to other languages, I think would help solve some debugging issues we've had around callbacks blocking/taking a long time to execute

lib/java/src/test/java/com/workiva/frugal/transport/FJmsSubscriberTransportTest.java
```diff
@@ -0,0 +1,64 @@
+package com.workiva.frugal.transport;
```
brianshannan-wf: Need to add license to all the java files

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def get_transport(self) -> FPublisherTransport:
+        """
+        Get a new FStompPublisherTransport.
+        """
+        return FStompPublisherTransport(
+            self._stomp_client, self._topic_prefix, self._max_message_size)
+
+
+class FStompPublisherTransport(FPublisherTransport):
```
brianshannan-wf: Can we make this class private? I think it would be better if that was the case

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,192 @@
...
+                 max_message_size: int=32 * 1024 * 1024):
+        super().__init__(max_message_size)
+        self._stomp_client = stomp_client
+        self._topic_prefix = topic_prefix
+
+    async def open(self):
+        """
+        No-op. Client connection should be established outside of frugal.
+        """
+        pass
```
brianshannan-wf: IMO we should make this set a dummy `_is_open` variable and enforce that. We've had problems in the past with vessel where we had these no-op methods, then needed them to do something at a later date. As people had been successfully using the class without opening it, we had a long discussion about whether that was a breaking change. Enforcing open is called with a dummy variable would prevent that.

Other opinions welcome though

lib/go/stomp_transport.go
blakeroberts-wk: Only invocation is [`FStompPublisherTransportFactory.GetTransport()`](https://github.com/Workiva/frugal/pull/1071/files#diff-3146009b39635b4a03b125b74eabf702R40).
`GetTransport()` is what consumers should *always* use, correct?
So we shouldn't export `NewStompPublisherTransport`

lib/python/frugal/transport/scope_transport.py
```diff
@@ -54,7 +54,7 @@ def publish(self, topic, data):
 
     def _check_publish_size(self, data):
         """
-        Returns True if the data is of permissible size, False otherwise.
+        Returns True if the length of data is over the max message size.
         """
         return len(data) > self._max_message_size > 0
```
brianshannan-wf: This is subtly wrong, I'll get it fixed while it's on my mind

scripts/skynet/cross/setup_java.sh
```diff
@@ -9,7 +9,7 @@ if [ -z "${IN_SKYNET_CLI+yes}" ]; then
 else
     cd ${FRUGAL_HOME}/lib/java
     mvn clean verify -q
-    mv target/frugal-2.20.0.jar ${FRUGAL_HOME}/test/integration/java/frugal-integration-test/frugal.jar
+    mv target/frugal-LOCAL.jar ${FRUGAL_HOME}/test/integration/java/frugal-integration-test/frugal.jar
```
brianshannan-wf: Should be changed

scripts/skynet/cross/setup_python.sh
```diff
@@ -13,3 +13,4 @@ fi
 pip install -e ".[tornado]"
 python3 /usr/bin/pip3 install Cython==0.27.2
 python3 /usr/bin/pip3 install -e ".[asyncio]"
+python3 /usr/bin/pip3 install aiostomp==1.4.0
```
brianshannan-wf: I don't think we need to do this? It should be covered by the previous line

test/integration/crossrunner/load.go
```diff
@@ -75,49 +78,76 @@ func newPair(client, server Config) *Pair {
...
 
 		// Append language level transports and protocols to client/server level
-		test.Client.Transports = append(test.Client.Transports, test.Transports...)
-		test.Server.Transports = append(test.Server.Transports, test.Transports...)
+		test.Client.Transports = append(test.Client.Transports, test.RpcTransports...)
+		test.Server.Transports = append(test.Server.Transports, test.RpcTransports...)
 		test.Client.Protocols = append(test.Client.Protocols, test.Protocols...)
 		test.Server.Protocols = append(test.Server.Protocols, test.Protocols...)
 
+		//test.Publisher.Transports = append(test.Publisher.Transports, test.Transports...)
```
brianshannan-wf: Remove this

test/integration/crossrunner/load.go
```diff
@@ -75,49 +78,76 @@ func newPair(client, server Config) *Pair {
...
+	}
+
+	pubsubPairs := make([]*Pair, 0)
+	println("len(publishers) == ", len(publishers))
+	println("len(subscribers) == ", len(subscribers))
+	for _, publisher := range publishers {
+		for _, subscriber := range subscribers {
+			println(publisher.Transport, subscriber.Transport, publisher.Protocol, subscriber.Protocol)
+			if publisher.Transport == subscriber.Transport && publisher.Protocol == subscriber.Protocol {
+				println("found a match!")
```
brianshannan-wf: found a match

lib/python/frugal/aio/transport/stomp_transport.py
blakeroberts-wk: It wouldn't break compilation, but it would break functionality; therefore, it's a breaking change. I think `_is_open` is a good idea.

lib/python/frugal/transport/scope_transport.py
blakeroberts-wk: Lol. Shouldn't we make `True` a good thing and `False` a bad thing. So like where this is used should flip too.

test/integration/tests.json
```diff
@@ -23,21 +23,45 @@
...
+    "rpcTransports": [ "nats", "http" ],
     "protocols": ["binary", "compact", "json"]
   },
   {
     "name": "java",
     "client": {
       "timeout": 20,
       "command": [ "/usr/bin/java", "-Djava.security.egd=file:///dev/urandom", "-cp", "cross.jar", "com.workiva.TestClient" ],
-      "transports": [ "nats", "http" ]
+      "transports": [ "nats", "http", "activemq" ]
```
brianshannan-wf: I don't think "activemq" should be added here

lib/python/frugal/transport/scope_transport.py
yuanmwang-wf: ^My thoughts too 😄 Especially since it's named `check_publish_size`, I'd assume `True` means good. Is it too late to flip now (will it be a breaking change)?

blakeroberts-wk: It shouldn't be a breaking change since it is a private method. It can only be invoked within this project (not literally, because python but still) so wherever it's invoked can be flipped without changing public api

brianshannan-wf: It's used by both the messaging-sdk and frugal though, so I would rather leave the return value the same

blakeroberts-wk: two invocation:
- https://github.com/Workiva/frugal/blob/master/lib/python/frugal/tornado/transport/nats_scope_transport.py#L66
- https://github.com/Workiva/frugal/blob/master/lib/python/frugal/aio/transport/nats_scope_transport.py#L88

lib/go/stomp_transport.go
yuanmwang-wf: Let's do it then to be keep it extensible for the future?

lib/go/stomp_transport.go
yuanmwang-wf: Agreed, I don't think it needs to be public, assuming consumers should always use `GetTransport`.

lib/java/src/main/java/com/workiva/frugal/transport/FJmsSubscriberTransport.java
yuanmwang-wf: 👍

scripts/skynet/cross/setup_java.sh
brianshannan-wf: Actually this is only when running skynet-cli locally, it doesn't really matter

examples/java/src/main/java/examples/ActiveMQPublisher.java
blakeroberts-wk:
```suggestion
import org.slf4j.Logger;
```

tylerrinnan-wf: It doesn't look like we're logging anything in the example?

lib/go/stomp_transport.go
```diff
@@ -90,54 +128,95 @@ func (m *fStompPublisherTransport) Publish(topic string, data []byte) error {
...
+	conn        *stomp.Conn
+	topicPrefix string
+	useQueue    bool
+}
+
+// newFStompSubscriberTransportFactory creates fStompSubscriberTransportFactory with the given stomp
 // connection and consumer name.
-func NewFStompSubscriberTransportFactory(conn *stomp.Conn, consumerPrefix string, useQueue bool) *FStompSubscribeTransportFactory {
-	return &FStompSubscribeTransportFactory{conn: conn, consumerPrefix: consumerPrefix, useQueue: useQueue}
+func newFStompSubscriberTransportFactory(conn *stomp.Conn, consumerPrefix string, useQueue bool) *fStompSubscriberTransportFactory {
```
blakeroberts-wk: nit: should the variable name be `topicPrefix` as well

lib/go/stomp_transport.go
```diff
@@ -156,9 +235,9 @@ func (m *fStompSubscriberTransport) Subscribe(topic string, callback FAsyncCallb
 
 	var destination string
 	if m.useQueue {
-		destination = fmt.Sprintf("/queue/%s%s%s", m.consumerPrefix, frugalPrefix, topic)
+		destination = fmt.Sprintf("/queue/%s%s%s", m.topicPrefix, frugalPrefix, topic)
```
blakeroberts-wk: It's an fqtn, a fully qualified topic name

lib/go/stomp_transport.go
```diff
@@ -22,6 +22,9 @@ import (
 	"github.com/go-stomp/stomp"
 )
 
+
+const defaultMaxPublishSize int = 32 * 1024 * 1024
```
yuanmwang-wf: nit: should this name mention stomp or mq?

yuanmwang-wf:
+1 

I think this is in a pretty good shape now, should we consider getting this released soon?

yuanmwang-wf:
Good catch @tylerrinnan-wf

lib/go/stomp_transport.go
blakeroberts-wk: I think we should since the const is in the generic frugal package

lib/go/stomp_transport.go
```diff
@@ -0,0 +1,324 @@
...
+			TRANSPORT_EXCEPTION_REQUEST_TOO_LARGE,
+			fmt.Sprintf("Message exceeds %d bytes, was %d bytes", m.maxPublishSize, len(data)))
+	}
+
+	destination := m.formatStompPublishTopic(topic)
+	logger().Debugf("frugal: publishing stomp message on topic '%s'", destination)
+	if err := m.conn.Send(destination, "application/octet-stream", data, stomp.SendOpt.Header("persistent", "true")); err != nil {
+		return thrift.NewTTransportExceptionFromError(err)
+	}
+	logger().Debugf("frugal: finished publishing stomp message to '%s'", destination)
```
blakeroberts-wk: Do we have any interest in recording publish times/durations?

blakeroberts-wk: nvm these are debug logs in a library

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @Override
+    public synchronized void close() {
+        if (!isOpen()) {
+            LOGGER.info("jms transport already closed, returning");
+            return;
+        }
+
+        try (Session closeSession = session) {
```
blakeroberts-wk: Is there a reason behind using `AutoClosable` for session and not producer?

blakeroberts-wk: nit because I'm guessing it doesn't matter.

But ... does this work?

```java
try (
    Session closeSession = session;
    MessageProducer closeProducer = producer
) {} catch (JMSException e) {
    LOGGER.error("failed to close jms producer", e);
} finally {
    session = null;
    producer = null;
}
```
👀

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
```diff
@@ -0,0 +1,191 @@
...
+    public synchronized void publish(String topic, byte[] payload) throws TTransportException {
+        if (!isOpen()) {
+            throw new TTransportException(TTransportException.NOT_OPEN, "failed to publish, jms client not open");
+        }
+
+        if (topic == null || "".equals(topic)) {
+            throw new TTransportException("publish topic cannot be empty");
+        }
+
+        if (payload.length > getPublishSizeLimit()) {
```
blakeroberts-wk:
```suggestion
        if (payload.length > getPublishSizeLimit() && getPublishSizeLimit() > 0) {
```

lib/python/frugal/aio/transport/stomp_transport.py
```diff
@@ -0,0 +1,242 @@
...
+
+    async def publish(self, topic: str, data):
+        """
+        Publish a message to stomp broker on a given topic.
+
+        Args:
+            topic: string
+            data: bytearray
+        """
+        if self._check_publish_size(data):
```
blakeroberts-wk: Same change about allowing 0 as other languages

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
brianshannan-wf: The point of using auto closeable is if the producer fails to close, we still want to close the session. I don't think using the auto closeable syntax for both matters, as the way it's written close will be called on both. I think the current way it's written is more readable, but that might just be me

lib/go/transport.go
```diff
@@ -40,8 +40,7 @@ type FPublisherTransport interface {
 	IsOpen() bool
 
 	// GetPublishSizeLimit returns the maximum allowable size of a payload
-	// to be published. A non-positive number is returned to indicate an
-	// unbounded allowable size.
+	// to be published. 0 is returned to indicate an unbounded allowable size.
```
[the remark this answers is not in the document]
brianshannan-wf: Yeah I think that's a good idea, I'll get that done

lib/go/stomp_transport.go
brianshannan-wf: So my intention with these debug logs is to allow users to more easily debug low level problems and rule out problems with sending messages, if that is the problem. I'm not sure if publish time is appropriate here, as that seems more high level telemetry data you'd want all the time rather than "turn on debug logging for a bit so we get more data". But it is a good point and I'm not sure what all should be here

lib/python/frugal/aio/transport/stomp_transport.py
brianshannan-wf: This actually already does that

lib/java/src/main/java/com/workiva/frugal/transport/FJmsPublisherTransport.java
blakeroberts-wk: Fair. And I mean readability is six of one, half a dozen of the other imo

blakeroberts-wk:
+1

test/integration/crossrunner/load.go
```diff
@@ -75,49 +78,70 @@ func newPair(client, server Config) *Pair {
...
 	bytes, err := ioutil.ReadFile(jsonFile)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	var tests []languages
 
 	// Unmarshal json into defined structs
+	println(jsonFile)
```
brianshannan-wf: remove

blakeroberts-wk:
+1

kevinfrommelt-wf:
+1

yuanmwang-wf:
@Workiva/release-management-p

PR #1077, opened 2019-01-29
Add reconnect handler for stomp transports

yuanmwang-wf:
### Story:
Adds reconnect handler for both stomp publisher and subscriber so it can be called within frugal when the original stomp connection is lost. 

### Acceptance Criteria:

### Design Notes:
TODO

### How To Test:
TODO

### My Test Results:
TODO

#### Reviewers:
@Workiva/product2


aviary-wf:
## Security Insights

## Action Items
---

yuanmwang-wf:
This is needed so reconnect logic can be implemented in messaging-sdk and passed to frugal. 

Confirmed it works locally if I stop and restart broker: https://screencast.com/t/HCmHf8l64Yp

lib/go/stomp_transport.go
```diff
@@ -22,22 +22,27 @@ import (
 	"github.com/go-stomp/stomp"
 )
 
+const (
+	defaultMaxPublishSize int = 32 * 1024 * 1024
+	stompClosed               = "connection already closed"
```
brianshannan-wf: I think we can reference this directly: https://github.com/go-stomp/stomp/blob/ba860900ef3307ecfe85ca14bc4838f24cfc7126/errors.go#L18

This way if that string is changed in the future our code will still work

yuanmwang-wf: Nice, will change 👍

lib/go/stomp_transport.go
```diff
@@ -131,11 +144,21 @@ func (m *fStompPublisherTransport) Publish(topic string, data []byte) error {
 	}
 
 	destination := m.formatStompPublishTopic(topic)
-	logger().Debugf("frugal: publishing stomp message on topic '%s'", destination)
 	if err := m.conn.Send(destination, "application/octet-stream", data, stomp.SendOpt.Header("persistent", "true")); err != nil {
+		if err.Error() == stompClosed {
+			e, conn := m.reconHandler()
```
brianshannan-wf: I think we need some sort of mutex here, otherwise we could have multiple go routines attempt to make this change at the same time

lib/go/stomp_transport.go
```diff
@@ -131,11 +144,21 @@ func (m *fStompPublisherTransport) Publish(topic string, data []byte) error {
 	}
 
 	destination := m.formatStompPublishTopic(topic)
-	logger().Debugf("frugal: publishing stomp message on topic '%s'", destination)
 	if err := m.conn.Send(destination, "application/octet-stream", data, stomp.SendOpt.Header("persistent", "true")); err != nil {
+		if err.Error() == stompClosed {
```
brianshannan-wf: Do we also need to check for https://github.com/go-stomp/stomp/blob/ba860900ef3307ecfe85ca14bc4838f24cfc7126/errors.go#L17 ? I noticed it was returned a few times in the connection code?

brianshannan-wf: We might want to lock here and check our open flag as well. The scenario I'm thinking of is
1. thread 1 starts `fStompPublisherTransport.Publish()`, but pauses before the `m.conn.Send()` call
2. thread 2 closes the connection and the transport
3. thread 1 starts again, and sees the connection is now closed, so it reconnects it

Where is this case, we wouldn't want that connection to be created again

lib/go/stomp_transport.go
```diff
@@ -283,22 +320,50 @@ func (m *fStompSubscriberTransport) Unsubscribe() error {
...
 		select {
 		case <-stopC:
-			logger().Errorf("frugal: error processing stomp subscription messages, message received on stop channel")
+			logger().Infof("frugal: received signal from stop channel, stop processing messages")
 			return
 		case message, ok := <-m.sub.C:
 			logger().Debugf("frugal: received stomp message on topic '%s'", m.topic)
 			if !ok {
-				logger().Errorf("frugal: error processing subscription messages, message channel closed")
+				logger().Warnf("frugal: message channel closed unexpectedly, will try to reconnect")
```
brianshannan-wf: I think it would be good to make sure this transport should be subscribed before reconnecting as well

lib/go/stomp_transport.go
brianshannan-wf: It would also be nice to have a unit test for this. Something like mocking out the stomp connection so it returns an error, and making sure a mock reconnect handler is called successfully and the publish is attempted again

lib/go/stomp_transport.go
```diff
@@ -127,11 +147,29 @@ func (m *fStompPublisherTransport) Publish(topic string, data []byte) error {
 	}
 
 	destination := m.formatStompPublishTopic(topic)
-	logger().Debugf("frugal: publishing stomp message on topic '%s'", destination)
 	if err := m.conn.Send(destination, "application/octet-stream", data, stomp.SendOpt.Header("persistent", "true")); err != nil {
+		if err == stomp.ErrAlreadyClosed || err == stomp.ErrClosedUnexpectedly {
+			// transport already closed, no reconnect should be attempted
+			if !m.isOpen {
```
blakeroberts-wk: Is it possible for the stomp connection to be already closed but the transport to consider itself open? I feel like this would mean there is a bug in our code that we should consider logging

lib/go/stomp_transport.go
```diff
@@ -279,19 +343,49 @@ func (m *fStompSubscriberTransport) Unsubscribe() error {
 	return nil
 }
 
+// reconnect calls reconHandler to try to get a new stomp connection and subscribe again.
+func (m *fStompSubscriberTransport) reconnect() error {
+	m.openMu.Lock()
+	err, conn := m.reconHandler(m.maxReconnAttempt)
```
blakeroberts-wk: Since it is the consumer that does the actual reconnecting (and has control over stuff like number of attempts and duration between consecutive attempts, etc) do we need the `maxReconnAttempt` parameter?

blakeroberts-wk: Like if we are going to take information about max reconnection attempts, should we maybe just take an address and tcp options and actually implement the reconnect handler ourselves?

yuanmwang-wf: The `maxReconnAttempt` can be configured in messaging-sdk can then propagate to frugal so consumers can easily set retry attempts if they use messaging-sdk: https://github.com/Workiva/messaging-sdk/blob/88b806041c94f6dfff2060bf124ff179d6979964/lib/go/sdk/active_mq_client.go#L267

yuanmwang-wf: I think taking addresses and tcp options would make this easier to read for sure (I actually did that in first pass), but I think the pattern of frugal transport is frugal should assume there's an active connection and any connection lifecycle related logic should be handled in messaging-sdk, so only exposing a reconnection method and have it implemented in messaging-sdk seems the most fitting to me here.

lib/go/stomp_transport.go
yuanmwang-wf: I'm not sure if it's possible for that to happen, but I'll add some logging here to be safe.

lib/go/stomp_transport.go
blakeroberts-wk: What I guess I'm trying to get at is `maxReconnAttempt` could actually be a `const` in the messaging sdk, and then we wouldn't need to have it mentioned in frugal
https://github.com/Workiva/messaging-sdk/blob/88b806041c94f6dfff2060bf124ff179d6979964/lib/go/sdk/active_mq_client.go#L166

But if we do want it in frugal, it would make since to also configure the duration between reconnect attempts
https://github.com/Workiva/messaging-sdk/blob/88b806041c94f6dfff2060bf124ff179d6979964/lib/go/sdk/active_mq_client.go#L191

Max retry attempts and duration between retries are both configurations to connection retries in general. If we would like to configure one retry option in frugal, shouldn't we configure all the options? But at the same time, because frugal doesn't actually use these configuration options, I feel like frugal shouldn't know about them at all. Like have the retry handler type be a function that takes no parameters

lib/go/stomp_transport.go
```diff
@@ -127,11 +147,29 @@ func (m *fStompPublisherTransport) Publish(topic string, data []byte) error {
...
 	destination := m.formatStompPublishTopic(topic)
-	logger().Debugf("frugal: publishing stomp message on topic '%s'", destination)
 	if err := m.conn.Send(destination, "application/octet-stream", data, stomp.SendOpt.Header("persistent", "true")); err != nil {
+		if err == stomp.ErrAlreadyClosed || err == stomp.ErrClosedUnexpectedly {
+			// transport already closed, no reconnect should be attempted
+			if !m.isOpen {
+				return nil
+			}
+
+			m.reconMu.Lock()
```
brianshannan-wf: If this transport is being used in multiple threads, I think this this reconnect can happen too many times. If multiple go routines get an error, they'd all block on this, then they'd all reconnect one by one, even though the connection was already reconnected

lib/go/stomp_transport.go
brianshannan-wf: I agree with Blake, the callback passed in could easily know the max reconnect number on its own. Passing in the current reconnect number (first reconnect, second reconnect) would make sense I think

yuanmwang-wf: Sounds good 👍

lib/go/stomp_transport.go
```diff
@@ -279,19 +343,49 @@ func (m *fStompSubscriberTransport) Unsubscribe() error {
...
+
 // processMessages call the given FAsyncCallback with messages from the
 // subscription channel.
 func (m *fStompSubscriberTransport) processMessages() {
 	stopC := m.stopC
 	for {
 		select {
 		case <-stopC:
-			logger().Errorf("frugal: error processing stomp subscription messages, message received on stop channel")
+			logger().Infof("frugal: received signal from stop channel, stop processing messages")
```
brianshannan-wf: I guess I'm not super clear under what conditions this can occur in. Can this happen as part of an unexpected disconnect?

yuanmwang-wf: I haven't seen this happen while testing, but I know `stopC` is closed when the transport [unsubscribes](https://github.com/Workiva/frugal/blob/05e57381935229a0439fe056b1ea3c6068f82bd2/lib/go/stomp_transport.go#L336), and I don't think that should be an error level log?

brianshannan-wf: Might be worth verifying that's the only case. If this can also be called for an unexpected disconnect then info seems low

yuanmwang-wf: Will do 👍

yuanmwang-wf:
going to close this for now

PR #1079, opened 2019-02-08
bump nats core pool size

brianshannan-wf:
Bump the core thread pool size for the default executor service for a FNatsServer. This means the default executor service will always have the maximum number of threads, but this trade off seems worth it given how thread pool executors scale threads. The executor service doesn't scale up threads until there are `queueLength` items waiting to be processed, which is very undesirable. I don't think there are many cases where having less than the maximum number of threads is desired.

A lot of teams currently supply their own executor service, but this should help teams who use the default one.

@Workiva/product2 @brettkail-wk

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -194,7 +194,7 @@ public class FNatsServer implements FServer {
         public FNatsServer build() {
             if (executorService == null) {
                 this.executorService = new ThreadPoolExecutor(
-                        1, workerCount, 30, TimeUnit.SECONDS,
+                        workerCount, workerCount, 30, TimeUnit.SECONDS,
                         new ArrayBlockingQueue<>(queueLength),
                         new BlockingRejectedExecutionHandler());
             }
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -194,7 +194,7 @@ public Builder withHighWatermark(long highWatermark) {
         public FNatsServer build() {
             if (executorService == null) {
                 this.executorService = new ThreadPoolExecutor(
-                        1, workerCount, 30, TimeUnit.SECONDS,
+                        workerCount, workerCount, 30, TimeUnit.SECONDS,
```
brettkail-wk: Update [doc](https://github.com/Workiva/frugal/pull/1079/files#diff-5786176bf915cb610c5fb4176826cfe9R160).

brettkail-wk: Since core is now max, keep alive time is irrelevant, so I would change to `0, TimeUnit.MILLISECONDS` for clarity, which would be consistent with `Executors.newFixedThreadPool`.  A small comment stating this default executor matches `newFixedThreadPool` but with different rejected execution handler might be helpful.

brianshannan-wf: Nice, forgot that was explicitly called out there

yuanmwang-wf:
+1

yuanmwang-wf:
@Workiva/release-management-p

PR #1081, opened 2019-02-11
go/java server event handlers

brianshannan-wf:
https://jira.atl.workiva.net/browse/INFENG-7385

This pull request does two, related, things.
1. Adds the concept of an `ephemeralProperties`, which is are key-value pairs that don't get serialized with the FContext. This allows information to be passed without it being serialized.
2. Adds an interface with with hooks for the `FNatsServer`. This interface has three methods
  - `onRequestReceived` - When the message is received from nats, but before it is placed on a queue for the processor to pick up.
  - `onRequestStarted` - When the processor starts processing the message
  - `onRequestFinished` - When the processor finishes processing the message

For go, I had the server take individual functions instead of an interface, as an interface in go can't have a default implementation, so we couldn't add any methods without it being a breaking change.

I believe this will expose enough functionality to allow more monitoring of the nats server, and extensible enough we can add more later.

Could be used to resolve https://github.com/Workiva/frugal/issues/798

@Workiva/product2 
@brettkail-wk @jesseregier-wf @tomdeering-wf I think you have all expressed interest in something like this in the past

### DIFF
--- a/lib/go/context.go
+++ b/lib/go/context.go
@@ -104,7 +104,7 @@ type FContextWithEphemeralProperties interface {
 
 	// Clone performs a deep copy of an FContextWithEphemeralProperties while
 	// handling opids correctly.
-	Clone() FContext
+	Clone() FContextWithEphemeralProperties
 
 	// EphemeralProperty gets the property associated with the given key.
 	EphemeralProperty(key interface{}) (interface{}, bool)
@@ -253,7 +253,9 @@ func (c *FContextImpl) Timeout() time.Duration {
 	return time.Millisecond * time.Duration(timeoutMillis)
 }
 
-func (c *FContextImpl) Clone() FContext {
+// Clone performs a deep copy of an FContextWithEphemeralProperties while
+// handling opids correctly.
+func (c *FContextImpl) Clone() FContextWithEphemeralProperties {
 	cloned := &FContextImpl{
 		requestHeaders: c.RequestHeaders(),
 		responseHeaders: c.ResponseHeaders(),
@@ -263,6 +265,7 @@ func (c *FContextImpl) Clone() FContext {
 	return cloned
 }
 
+// EphemeralProperty gets the property associated with the given key.
 func (c *FContextImpl) EphemeralProperty(key interface{}) (interface{}, bool) {
 	c.mu.Lock()
 	value, ok := c.ephemeralProperties[key]
@@ -270,6 +273,7 @@ func (c *FContextImpl) EphemeralProperty(key interface{}) (interface{}, bool) {
 	return value, ok
 }
 
+// EphemeralProperties returns a copy of the ephemeral properties map.
 func (c *FContextImpl) EphemeralProperties() map[interface{}]interface{} {
 	c.mu.RLock()
 	properties := make(map[interface{}]interface{}, len(c.ephemeralProperties))
@@ -280,6 +284,7 @@ func (c *FContextImpl) EphemeralProperties() map[interface{}]interface{} {
 	return properties
 }
 
+// AddEphemeralProperty adds a keyp-value pair to the ephemeral properties.
 func (c *FContextImpl) AddEphemeralProperty(key, value interface{}) FContext {
 	c.mu.Lock()
 	c.ephemeralProperties[key] = value
--- a/lib/go/context.go
+++ b/lib/go/context.go
@@ -95,13 +95,40 @@ type FContext interface {
 	Timeout() time.Duration
 }
 
+// FContextWithEphemeralProperties is an extension of the FContext interface
+// with support for ephemeral properties. Ephemeral properties are a map of
+// key-value pairs that won't be serialized with the rest of the FContext.
+// TODO 4.0 add this to the FContext interface
+type FContextWithEphemeralProperties interface {
+	FContext
+
+	// Clone performs a deep copy of an FContextWithEphemeralProperties while
+	// handling opids correctly.
+	Clone() FContext
+
+	// EphemeralProperty gets the property associated with the given key.
+	EphemeralProperty(key interface{}) (interface{}, bool)
+
+	// EphemeralProperties returns a copy of the ephemeral properties map.
+	EphemeralProperties() map[interface{}]interface{}
+
+	// AddEphemeralProperty adds a keyp-value pair to the ephemeral properties.
+	AddEphemeralProperty(key, value interface{}) FContext
+}
+
 // Clone performs a deep copy of an FContext while handling opids correctly.
-// TODO 3.0 consider adding this to the FContext interface.
+// TODO 4.0 consider adding this to the FContext interface.
 func Clone(ctx FContext) FContext {
+	if fctxWEP, ok := ctx.(FContextWithEphemeralProperties); ok {
+		return fctxWEP.Clone()
+	}
+
 	clone := &FContextImpl{
 		requestHeaders:  ctx.RequestHeaders(),
 		responseHeaders: ctx.ResponseHeaders(),
+		ephemeralProperties: make(map[interface{}]interface{}),
 	}
+
 	clone.requestHeaders[opIDHeader] = getNextOpID()
 	return clone
 }
@@ -114,9 +141,10 @@ func getNextOpID() string {
 
 // FContextImpl is an implementation of FContext.
 type FContextImpl struct {
-	requestHeaders  map[string]string
-	responseHeaders map[string]string
-	mu              sync.RWMutex
+	requestHeaders      map[string]string
+	responseHeaders     map[string]string
+	ephemeralProperties map[interface{}]interface{}
+	mu                  sync.RWMutex
 }
 
 // NewFContext returns a Context for the given correlation id. If an empty
@@ -223,6 +251,40 @@ func (c *FContextImpl) Timeout() time.Duration {
 		return defaultTimeout
 	}
 	return time.Millisecond * time.Duration(timeoutMillis)
+}
+
+func (c *FContextImpl) Clone() FContext {
+	cloned := &FContextImpl{
+		requestHeaders: c.RequestHeaders(),
+		responseHeaders: c.ResponseHeaders(),
+		ephemeralProperties: c.EphemeralProperties(),
+	}
+	cloned.requestHeaders[opIDHeader] = getNextOpID()
+	return cloned
+}
+
+func (c *FContextImpl) EphemeralProperty(key interface{}) (interface{}, bool) {
+	c.mu.Lock()
+	value, ok := c.ephemeralProperties[key]
+	c.mu.Unlock()
+	return value, ok
+}
+
+func (c *FContextImpl) EphemeralProperties() map[interface{}]interface{} {
+	c.mu.RLock()
+	properties := make(map[interface{}]interface{}, len(c.ephemeralProperties))
+	for key, value := range c.ephemeralProperties {
+		properties[key] = value
+	}
+	c.mu.RUnlock()
+	return properties
+}
+
+func (c *FContextImpl) AddEphemeralProperty(key, value interface{}) FContext {
+	c.mu.Lock()
+	c.ephemeralProperties[key] = value
+	c.mu.Unlock()
+	return c
 }
 
 // setRequestOpID sets the request operation id for context.
--- a/lib/java/src/main/java/com/workiva/frugal/FContext.java
+++ b/lib/java/src/main/java/com/workiva/frugal/FContext.java
@@ -257,6 +257,8 @@ public class FContext implements Cloneable {
 
     /**
      * Sets the given map as the ephemeral properties for the FContext.
+     * A reference to the map will be kept, so modifications made to the map
+     * outside this object will affect the ephemeral properties in this object.
      *
      * @param ephemeralProperties map to set as ephemeral properties
      * @return FContext
@@ -287,12 +289,12 @@ public class FContext implements Cloneable {
      * property with the same name is already present, it will be replaced. The same
      * FContext is returned to allow for call chaining.
      *
-     * @param ephemeralProperties ephemeral properties to add
+     * @param properties ephemeral properties to add
      * @return FContext
      */
-    public FContext addEphemeralProperties(Map<Object, Object> ephemeralProperties) {
-        for (Map.Entry<Object, Object> pair : ephemeralProperties.entrySet()) {
-            addEphemeralProperty(pair.getKey(), pair.getValue());
+    public FContext addEphemeralProperties(Map<Object, Object> properties) {
+        for (Map.Entry<Object, Object> pair : properties.entrySet()) {
+            ephemeralProperties.putAll(properties);
         }
         return this;
     }
--- a/lib/java/src/main/java/com/workiva/frugal/FContext.java
+++ b/lib/java/src/main/java/com/workiva/frugal/FContext.java
@@ -24,6 +24,8 @@ import java.util.concurrent.atomic.AtomicLong;
  * can be used to set request headers, response headers, and the request timeout.
  * The default timeout is five seconds. An FContext is also sent with every publish
  * message which is then received by subscribers.
+ * An FContext can also contain a map of properties that won't be serialized,
+ * called ephemeralProperties.
  * <p>
  * In addition to headers, the FContext also contains a correlation ID which can
  * be used for distributed tracing purposes. A random correlation ID is generated
@@ -66,6 +68,7 @@ public class FContext implements Cloneable {
 
     private Map<String, String> requestHeaders = new ConcurrentHashMap<>();
     private Map<String, String> responseHeaders = new ConcurrentHashMap<>();
+    private Map<Object, Object> ephemeralProperties = new HashMap<>();
 
     private FContext(Map<String, String> requestHeaders, Map<String, String> responseHeaders) {
         this.requestHeaders = requestHeaders;
@@ -253,6 +256,70 @@ public class FContext implements Cloneable {
     }
 
     /**
+     * Sets the given map as the ephemeral properties for the FContext.
+     *
+     * @param ephemeralProperties map to set as ephemeral properties
+     * @return FContext
+     */
+    public FContext setEphemeralProperties(Map<Object, Object> ephemeralProperties) {
+        this.ephemeralProperties = ephemeralProperties;
+        return this;
+    }
+
+    /**
+     * Adds an ephemeral property to the FContext. An ephemeral property is a key-value
+     * pair that will not be serialized with the FContext. If an ephemeral property
+     * with the same name is already present, it will be replaced. The same FContext
+     * is returned to allow for call chaining.
+     *
+     * @param key ephemeral property name
+     * @param value ephemeral property value
+     * @return FContext
+     */
+    public FContext addEphemeralProperty(Object key, Object value) {
+        ephemeralProperties.put(key, value);
+        return this;
+    }
+
+    /**
+     * Adds the given ephemeral properties to the FContext. An ephemeral property is
+     * a key-value pair that will not be serialized with the FContext. If an ephemeral
+     * property with the same name is already present, it will be replaced. The same
+     * FContext is returned to allow for call chaining.
+     *
+     * @param ephemeralProperties ephemeral properties to add
+     * @return FContext
+     */
+    public FContext addEphemeralProperties(Map<Object, Object> ephemeralProperties) {
+        for (Map.Entry<Object, Object> pair : ephemeralProperties.entrySet()) {
+            addEphemeralProperty(pair.getKey(), pair.getValue());
+        }
+        return this;
+    }
+
+    /**
+     * Returns the ephemeral properties on the FContext. Ephemeral properties are
+     * key-value pairs that will not be serialized with the FContext.
+     *
+     * @return ephemeral properties
+     */
+    public Map<Object, Object> getEphemeralProperties() {
+        return new HashMap<>(ephemeralProperties);
+    }
+
+    /**
+     * Returns the ephemeral property from the FContext for the corresponding key,
+     * Or null if it doesn't exist. An ephemeral property is a key-value pair that
+     * will not be serialized with the FContext.
+     *
+     * @param key ephemeral property name
+     * @return ephemeral property value
+     */
+    public Object getEphemeralProperty(Object key) {
+        return ephemeralProperties.get(key);
+    }
+
+    /**
      * Get the request timeout.
      *
      * @return the request timeout in milliseconds.
@@ -275,6 +342,7 @@ public class FContext implements Cloneable {
         FContext cloned = (FContext) super.clone();
         cloned.requestHeaders = this.getRequestHeaders();
         cloned.responseHeaders = this.getResponseHeaders();
+        cloned.ephemeralProperties = this.getEphemeralProperties();
         cloned.addRequestHeader(OPID_HEADER, getNextOpId());
         return cloned;
     }
--- a/lib/java/src/main/java/com/workiva/frugal/FContext.java
+++ b/lib/java/src/main/java/com/workiva/frugal/FContext.java
@@ -293,9 +293,7 @@ public class FContext implements Cloneable {
      * @return FContext
      */
     public FContext addEphemeralProperties(Map<Object, Object> properties) {
-        for (Map.Entry<Object, Object> pair : properties.entrySet()) {
-            ephemeralProperties.putAll(properties);
-        }
+        ephemeralProperties.putAll(properties);
         return this;
     }
 
--- a/lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
+++ b/lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
@@ -24,6 +24,7 @@ import org.apache.thrift.protocol.TSet;
 import org.apache.thrift.protocol.TStruct;
 
 import java.nio.ByteBuffer;
+import java.util.HashMap;
 import java.util.Map;
 
 import static com.workiva.frugal.FContext.CID_HEADER;
@@ -41,10 +42,16 @@ import static com.workiva.frugal.FContext.OPID_HEADER;
 public class FProtocol extends TProtocol {
 
     private TProtocol wrapped;
+    private Map<Object, Object> ephemeralProperties;
 
     protected FProtocol(TProtocol proto) {
         super(proto.getTransport());
         wrapped = proto;
+        ephemeralProperties = new HashMap<>();
+    }
+
+    public void setEphemeralProperties(Map<Object, Object> ephemeralProperties) {
+        this.ephemeralProperties = ephemeralProperties;
     }
 
     /**
@@ -77,6 +84,10 @@ public class FProtocol extends TProtocol {
         if (cid != null && !cid.isEmpty()) {
             ctx.addResponseHeader(CID_HEADER, cid);
         }
+
+        // setting the ephemeral properties allows this map in FContext and the
+        // map in the server to be the same.
+        ctx.setEphemeralProperties(ephemeralProperties);
         return ctx;
     }
 
--- a/lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
+++ b/lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
@@ -50,6 +50,12 @@ public class FProtocol extends TProtocol {
         ephemeralProperties = new HashMap<>();
     }
 
+    /**
+     * Sets a map of properties that should be set on an FContext read from
+     * {@link FProtocol#readRequestHeader()} as ephemeral properties.
+     *
+     * @param ephemeralProperties The properties to set on a FContext
+     */
     public void setEphemeralProperties(Map<Object, Object> ephemeralProperties) {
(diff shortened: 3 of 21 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -267,16 +283,26 @@ protected MessageHandler newRequestHandler() {
...
 
+            // Get the requests correlation id for more information for each request
+            String correlationId;
+            try {
+                correlationId = HeaderUtils.decodeFromFrame(message.getData()).get(FContext.CID_HEADER);
+            } catch (TException e) {
+                LOGGER.warn("Discarding invalid NATS request, could not get FContext headers", e);
+                return;
+            }
+            eventHandler.onNewRequest(correlationId);
```
brettkail-wk: It's unfortunate that we unconditionally deserialize all the headers here, especially since the default handler doesn't use the correlation ID.

I've not thought deeply, but it seems it would be better to somehow propagate this down to `FContext` so that all this watermark/logging logic could be done by a standard middleware (and optionally disabled in the library altogether).  For example, propagate the timestamp (in some form) to the inbound `FProtocol` so that `readRequestHeader` can propagate it to some kind of non-serialized attributes on `FContext`.

As for the timestamps themselves, `System.currentTimeMillis()` has the monotonic time issue that was resolved in Go `time.Time`, and it has granularity issues that are solved in `Instant` in Java 9.  Perhaps both `Instant` and monotonic time should be stored as non-serialized attribute (or some kind of opaque `FTimestamp` that unifies them as `time.Time` does).

brianshannan-wf: The problem with a combined `System.currentTimeMillis()` and `Instant` is FContext is typed as a `Map<String, String>` right now

brettkail-wk: The problem with using a header is that it propagates to downstream calls.  By "some king of non-serialized attributes", I meant adding some kind of `Map<String, Object> attributes` to `FContext` ala `ServletRequest`, which has both in-memory and serialized maps.  Adding another map would cause complexities for `clone`, which can perhaps be solved by documenting that it uses shallow clone only and that immutable types are encouraged.

In other places, we hack around the lack of non-serialized attributes by setting/unsetting _response_ headers in middleware.

Note: `System.currentTimeMillis` is approximately equivalent to `Instant`, which is the system clock that can be adjusted by ntpd, etc.  For timings, you want to compare `nanoTime`, the monotonic clock.  Go's `time.Time` combines both in a single object.

In any case, even if you use the response header trick, you still have to somehow serialize the system+monotonic values.  Perhaps you can come up with a string serialization and add helpers to `Headers` so consumers don't have to figure it out.

brianshannan-wf: Hmm yeah I hadn't considered FContext reuse. Okay, I'll see what I can do

brianshannan-wf: Okay, I added the concept of ephemeral headers the FNatsServer can get into the FContext headers. I made this a separate map from the other headers so they won't be copied on clone, and won't be serialized if the FContext is reused.

I think my main concern with this method is getting people to understand the difference between the two and which headers are in which map, but overall I think this is definitely a better approach

Thoughts?

jesseregier-wf:
>How useful are hooks for when a new request is received (before being put onto a work queue) and when a request is finished? Does a combination of tracing in middleware and a high watermark handler fulfill all the things a request received/finished handler does?

The reason I was looking for these in the past was so I could compensate for the missing telemetry/statistics data from Messaging SDK. I had been looking for a variety of stats, but IIRC, that "begin/end request" was what I would need to capture everything. If we have better/other ways to capture these details, there's no need for those hooks.

We were basically looking for ways to investigate and diagnose issues after the fact. This was our short-list:
- request latency (How long are tasks sitting in the queue before processing begins?)
- request queue depth/capacity (How many outstanding requests are waiting in the queue right now?)
- how many requests have been dropped? (Never added to the queue in the first place cause it's full)

We had potential workarounds for some of those (i.e. if we supplied our own `ThreadPoolExecutor` then we'd control the worker queue and could report stats on it ourselves) but that's really pretty unnecessary/over-complicated. Ideally, the SDK should support some of this without jumping through hoops.

I haven't had a chance to go through this thoroughly yet, but I think this would answer some of our questions (i.e. capturing that "received timestamp" allows us to record request processing latency in middleware)

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -252,6 +253,22 @@ public String getResponseHeader(String name) {
         return new HashMap<>(responseHeaders);
     }
 
+    public FContext addEphemeralHeader(String key, Object value) {
```
brettkail-wk: Quibbling, "header" sounds like something that is transferred, so "ephemeral header" isn't very meaningful.  Perhaps attribute, property, metadata, ..

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -66,6 +66,7 @@
 
     private Map<String, String> requestHeaders = new ConcurrentHashMap<>();
     private Map<String, String> responseHeaders = new ConcurrentHashMap<>();
+    private Map<String, Object> ephemeralHeaders = new HashMap<>();
```
brettkail-wk: Obviously still WIP, but seems unusual the others are (unexpectedly?) `ConcurrentHashMap` but not this one.

brianshannan-wf:
- I think passing the received timestamp through would satisfy the request latency piece. It wouldn't report anything automatically, but tracing doesn't happen automatically either.
- How would you want request queue depth to be reported? What value does this provide you if you know how long something waited in the queue? I'm not saying it's not, just curious as I had kind of equated those two things.
- By default an FNatsServer doesn't drop any requests, it blocks putting them into a queue, at least in go and java. This behaviour is customizable in java via executor service settings, but not in go. So I'm not sure how valuable that metric is, but I could be missing something

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -304,8 +322,7 @@ protected MessageHandler newRequestHandler() {
         public void run() {
             long duration = System.currentTimeMillis() - timestamp;
             if (duration > highWatermark) {
-                LOGGER.warn(String.format(
-                        "request spent %d ms in the transport buffer, your consumer might be backed up", duration));
+                eventHandler.onHighWatermark(duration);
```
brettkail-wk: Given the other changes, I wonder if the handler is needed at all or whether just need a way to disable it.  If keeping it, I wonder if `onReceived(Map<String, Object>)` and `onProcess(Map<String, Object>)`, and then all `_received_timestamp` and `highWatermark` logic could move into the handler.

lib/java/src/main/java/com/workiva/frugal/FContext.java
brianshannan-wf: We intended FContext to be threadsafe, so this should be a `ConcurrentHashMap`

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brianshannan-wf: Yeah that's fair, if a high watermark isn't provided (or is provided as zero) that could be the same as disabling. So `onReceived(Map<String, Object>)` and `onProcess(Map<String, Object>)` would be user provided callbacks that take the "ephemeral headers" (or whatever we call them)? If that's what you mean, I guess I'm not sure what benefit those provide over middleware

jesseregier-wf:
> I think passing the received timestamp through would satisfy the request latency piece. It wouldn't report anything automatically, but tracing doesn't happen automatically either.

Yep, that's all I'd need - middleware can worry about the rest of it.

> How would you want request queue depth to be reported?

I was going to use the start/end hooks to increment/decrement some counter that could be sampled for telemetry reporting.

> What value does this provide you if you know how long something waited in the queue? I'm not saying it's not, just curious as I had kind of equated those two things.

Generally, greater visibility, but they answer two different questions in my mind. The "high watermark" tells you that you're currently behind in processing, the number of outstanding tasks tells you how much further you have to go (one could argue this could **roughly** be calculated from the watermark time & average task time).

While under load, content-permissions spends all of it's time waiting on Eva, very little time is actually spent processing requests: the only way we'd ever scale on CPU usage is if we added too many worker threads and they were thrashing. Back when we were in Harbour, we'd hoped to use that count to intelligently fire CloudWatch alerts to force a new instance to scale in based on some computation (# of outstanding work, total # workers, average task time, etc) This idea wasn't even prototyped, so it's possible we could infer everything necessary from the task latency.

> By default an FNatsServer doesn't drop any requests, it blocks putting them into a queue, at least in go and java.

Good to know! That does sound vaguely familiar. (I'm pretty sure there can't be infinite blocked requests though, so *something* is getting dropped, somewhere.... but that's a whole different conversation).

lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
```diff
@@ -41,10 +42,21 @@
 public class FProtocol extends TProtocol {
 
     private TProtocol wrapped;
+    private Map<String, Object> ephemeralHeaders;
 
     protected FProtocol(TProtocol proto) {
         super(proto.getTransport());
         wrapped = proto;
+        ephemeralHeaders = new HashMap<>();
```
yuanwang-wf: does here need to be `ConcurrentHashMap` as well

lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
```diff
@@ -41,10 +42,21 @@
 public class FProtocol extends TProtocol {
 
     private TProtocol wrapped;
+    private Map<String, Object> ephemeralHeaders;
```
yuanwang-wf: does here need to be `ConcurrentHashMap` as well

brianshannan-wf: It doesn't, FProtocol isn't intended to be threadsafe

brianshannan-wf:
So for monitoring the queue size in java I'm currently of the opinion that providing your own `ThreadPoolExecutor` is the way to go. This is because:
- It provides an easy way to check the size of the queue. You wouldn't have to implement a stateful class to track queue/enqueues of the executor service
- You can implement your own `RejectedExecutionHandler` to do something when the queue is full

These points aren't super accurate if you're using a different kind of `ExecutorService`, and it doesn't really apply to go. So maybe those methods are necessary, and they probably have uses other than this particular case. But I'd prefer to only add methods we deem necessary as we can't take them away in the future, so I'd be curious why you'd prefer them over creating and monitoring a `ThreadPoolExecutor` yourself.

brianshannan-wf:
Was there any additional thoughts around needed features with this? Or consensus around if this is all that is currently needed?

brettkail-wk:
@brianshannan-wf I have not had time to catch up on this conversation, sorry.

lib/java/src/main/java/com/workiva/frugal/server/FServerEventHandler.java
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Provides an interface with which to handle events from an FServer.
+ */
+public interface FServerEventHandler {
+
+    /**
+     * Called when the FServer encounters a request that took an excess amount
+     * of time to begin processing.
+     */
+    void onHighWatermark(long duration);
```
brettkail-wk: This interface feels too specialized.  My suggestion is to make this interface be a server-level middleware to complement the existing service-level middleware:
```
interface FServerRequestEventHandler {
  void onRequestReceived(Map<Object, Object> ephemeralProperties);
  void onRequestStarted(Map<Object, Object> ephemeralProperties);
  void onRequestEnded(Map<Object, Object> ephemeralProperties);
}
```
With this, all of the current high watermark logic can be pushed into `FDefaultNatsServerHandler`:
```
public class FServerEventHandler {
  private long highWatermark;

  public void onRequestReceived(Map<Object, Object> ephemeralProperties) {
    ephemeralProperties.put("_received_nanotime", System.nanoTime());
  }

  public void onRequestReceived(Map<Object, Object> ephemeralProperties) {
    duration = System.nanoTime() - (long) ephemeralProperties.get("_received_nanotime");
    if (duration > highWatermark) { ... }
  }
}
```
This kind of interface would give a large amount of flexibility to clients of `FNatsServer`.

brettkail-wk:
> I think passing the received timestamp through would satisfy the request latency piece. It wouldn't report anything automatically, but tracing doesn't happen automatically either.

Well, `System.nanoTime` is the reliable clock for timing, so it seems weird to create an API that passes a timestamp intended for timing that doesn't use that.  That said, I appreciate `System.nanoTime` probably feels like an odd basis for an API.  Go handles this nicely with `time.Time`, which is somewhat like a `Instant.now()` (higher resolution than `System.currentTimeMillis()`) + `System.nanoTime()`, but Java doesn't have something like that, so we're in a somewhat odd position.  Perhaps it's not worth making the builtin timestamp visible at all (so continue to keep the watermark "internal"), but give services the hooks to implement their own mechanism.

> By default an FNatsServer doesn't drop any requests, it blocks putting them into a queue, at least in go and java.

The blocking queue is limited to 64 entries, and then the receiver thread blocks until it can insert into the buffer, which means it stops receiving new inbound messages from the NATS server until that happens.  Will the NATS server buffer messages indefinitely for a server, or does it drop them if the server stops receiving messages?

> Was there any additional thoughts around needed features with this? Or consensus around if this is all that is currently needed?

Our services use `FContext.getTimeout` to cooperatively abort in-progress requests if they detect they have run longer than the client-provided timeout, which avoids wasting server resources if the client isn't going to use the response anyway.  Currently, the only option for starting my timer is within the service method, which doesn't account for the time spent in the transport buffer.  If the transport buffer is backed up, then my server is overloaded, so it's even more important to proactively abort requests (possibly before they even start) to avoid using server resources to produce responses that will be ignored.

To do all that, I need to be able to capture a timestamp from the "request received" (either document `_received_nanotime`, or if we decide that's not worth the API complexity, then I can register my own event handler and store `System.nanoTime` myself), and I need to be able to propagate it via `FContext.  So for me, I'm hoping the outcome is:
1.  A hook so that services can override the builtin transport buffer warning logic with whatever sophistication they want.
2. A hook so that services can store non-serialized properties on FContext.  This would be useful for propagating timestamps from "request received", but it is generally useful.

lib/java/src/main/java/com/workiva/frugal/server/FServerEventHandler.java
brettkail-wk: I forgot to mention: perhaps this hook could allow us to fix the long-standing issue where our logging pipeline expect services to use the `"correlationId"` MDC key, but messages logged by `FBaseProcessor` don't have that.  With this hook in place, we could remove that MDC key from `onRequestEnded` if an ephemeral property indicates middleware stored the key...

brianshannan-wf: I think I agree with you. I was hesitant to have `onRequestStarted()`/`onRequestEnded()` methods as I thought it wouldn't provide much value over the current service level middleware and would confuse people as to which one they should use. But I think the APIs would be different enough that it should be somewhat clear, and would provide value service level middleware wouldn't

brianshannan-wf:
That makes sense. One of my remaining questions is do we want to expose transport level errors to callbacks in some way? Maybe a generic `onTransportError()`? Examples would be:
- https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java#L271
- https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java#L346

brettkail-wk:
@brianshannan-wf Sure, an error handler callback could be reasonable, so long as we don't mix the "scope" of callback (request-specific vs server-wide).  [This error handling](https://github.com/Workiva/frugal/blob/a10936b46d7935599779015ed6421c4ddc7ed535/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java#L333) also looks questionable.

yuanmwang-wf:
I think you need this to make workiva build pass: https://github.com/Workiva/frugal/pull/1110/commits/e333617a2d42dcb98e40a9023dece5900b731ab3

brianshannan-wf:
I think the ppa that is changing isn't needed, so I removed it and we'll see what happens. If it is needed for some reason I can add that flag

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -252,6 +253,26 @@ public String getResponseHeader(String name) {
         return new HashMap<>(responseHeaders);
     }
 
+    public FContext addEphemeralProperty(Object key, Object value) {
```
stevenosborne-wf: docs

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -66,6 +68,7 @@
 
     private Map<String, String> requestHeaders = new ConcurrentHashMap<>();
     private Map<String, String> responseHeaders = new ConcurrentHashMap<>();
+    private Map<Object, Object> ephemeralProperties = new HashMap<>();
```
brettkail-wk: If any of these maps should reasonably be `ConcurrentHashMap`, this seems like the one.  The others seem like they would be modified only at the edges of the service, but `ephemeralProperties` seems like it could be modified elsewhere.  Having one of the maps be different from the all the others seems risky, but if you prefer plain `HashMap`, perhaps make note of that in the javadoc?

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -252,6 +255,70 @@ public String getResponseHeader(String name) {
         return new HashMap<>(responseHeaders);
     }
 
+    /**
+     * Sets the given map as the ephemeral properties for the FContext.
```
brettkail-wk: Perhaps explicitly state that the `FContext` maintains a reference to the map and modifications to the map after the call will affect the ephemeral properties of the context object.

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -252,6 +255,70 @@ public String getResponseHeader(String name) {
...
+     * a key-value pair that will not be serialized with the FContext. If an ephemeral
+     * property with the same name is already present, it will be replaced. The same
+     * FContext is returned to allow for call chaining.
+     *
+     * @param ephemeralProperties ephemeral properties to add
+     * @return FContext
+     */
+    public FContext addEphemeralProperties(Map<Object, Object> ephemeralProperties) {
+        for (Map.Entry<Object, Object> pair : ephemeralProperties.entrySet()) {
+            addEphemeralProperty(pair.getKey(), pair.getValue());
```
brettkail-wk: Suggest `this.ephemeralProperties.putAll(ephemeralProperties)`.

lib/java/src/main/java/com/workiva/frugal/server/FDefaultNatsServerEventHandler.java
```diff
@@ -0,0 +1,44 @@
...
+    public void onRequestReceived(Map<Object, Object> ephemeralProperties) {
+        long now = clock.millis();
+        ephemeralProperties.put(REQUEST_RECEIVED_MILLIS_KEY, now);
+    }
+
+    @Override
+    public void onRequestStarted(Map<Object, Object> ephemeralProperties) {
+        if (ephemeralProperties.get(REQUEST_RECEIVED_MILLIS_KEY) != null) {
+            long started = (long) ephemeralProperties.get(REQUEST_RECEIVED_MILLIS_KEY);
+            long duration = System.currentTimeMillis() - started;
```
brettkail-wk: Seems like this should also use `clock`.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServerEventHandler.java
```diff
@@ -0,0 +1,31 @@
+package com.workiva.frugal.server;
+
+import java.util.Map;
+
+/**
+ * Provides an interface with which to handle events from an FNatsServer.
```
brettkail-wk: Perhaps highlight the differences from middleware here.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -349,9 +361,11 @@ protected MessageHandler newRequestHandler() {
                 return;
             }
 
+            Map<Object, Object> ephemeralProperties = new HashMap<>();
+            this.eventHandler.onRequestReceived(ephemeralProperties);
```
brettkail-wk: Can you remind the rationale for not updating HTTP?  It would be useful if we can rely on the same event handler for all inbound traffic.

lib/java/src/main/java/com/workiva/frugal/FContext.java
brianshannan-wf: This is a little confusing and made me take a second look though the code, even though the defaults for the other maps are `ConcurrentHashMap`, other map implementations can be passed in via [the headers constructor](https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/FContext.java#L99). In practice, the nats server always passes in a `HashMap`. In fairness, the class is documented as not [thread safe](https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/FContext.java#L38) but it's not super apparent. Additionally, all cloned FContexts will have `HashMaps` [for their header maps](https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/FContext.java#L276). I debated changing the other defaults, but leaned toward not doing that in case another server didn't pass in a thread safe map implementation and consumers relied on that.

I decided to go with a regular `HashMap` given this information, but I'm not convinced it's the best solution.

brettkail-wk: Thanks for all that context, that's really helpful.  Given all that, I think what you've done is quite reasonable.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brianshannan-wf: My concern is we may want to have different methods for HTTP.
- The HTTP server doesn't have the async executor service behaviour as the nats server, making the `onRequestReceived`/`onRequestStarted` methods redundant
- HTTP requests have much more information on them than nats messages, such as the headers. I was thinking it might be useful to expose the http request object to the event handler, so the handler could do things with them. A use case that has been discussed recently is getting the source ip from the headers for audit logging. Might not work out, but I thought leaving the option open was good

lib/go/context.go
```diff
@@ -95,13 +95,40 @@ type FContext interface {
 	Timeout() time.Duration
 }
 
+// FContextWithEphemeralProperties is an extension of the FContext interface
+// with support for ephemeral properties. Ephemeral properties are a map of
+// key-value pairs that won't be serialized with the rest of the FContext.
+// TODO 4.0 add this to the FContext interface
+type FContextWithEphemeralProperties interface {
```
brianshannan-wf: I'm not super happy about his, but it's the only way I could think of to do this

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brettkail-wk: The redundancy of `onRequestReceived` / `onRequestStarted` doesn't seem that bad, and in theory, the HTTP server could even manage the concurrency similarly to NATS (e.g., using async servlets and a custom thread pool).  That would allow the concurrency of Frugal requests to be managed differently from the concurrency of the HTTP server as a whole. Currently, there is no Frugal-level way of managing of the concurrency of individual HTTP like there is for NATS services.

I imagined that HTTP request/response headers could be specified as values of specific keys in the ephemeral properties so that they are accessible to anything holding the `FContext` (e.g., `FHttpRequest` as `_http_request`, etc.).

brianshannan-wf: I'm not completely sure how async servlets work, but yes that would be a possibility in the future.

I guess my concern with that approach is making the properties map a collection of objects transparently added by the server with no typing, and how that will look a year from now. It would be more flexible for adding new things than a different interface though, I'll think on that

brettkail-wk: Async servlet gives up the servlet thread before sending a response, and then gives an API to "complete" the response later in another thread.  Basically, it means we could "trade" a running servlet thread for a running Frugal processor thread, which is very similar to what we do for NATS today (we queue the work to run on another thread, then the NATS reader thread goes back to waiting to handle the next inbound message), which has the advantage of using a different set of threads.  (Of course, Go dodges this complexity of explicit async programming with runtime magic of goroutines :-).)

As for type-safety, this is basically the same as [ServletRequest.getAttribute](https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getAttribute(java.lang.String)).  Consumers and producers document the key name and expected type, and sometimes you write a wrapper/utility function to dig the attribute out (e.g., a static `FHttpRequest.get(FContext)` could be added).  Similar to `context.Context.Value` in Go.

Another approach I've seen is `public <T> void addAttribute(Class<T> key, T value)` and `public <T> T getAttribute(Class<T> key)` that uses `Class.cast` internally.  That makes the API a bit cleaner for code retrieving values but makes it a bit harder for code storing values (you always have to store a specific type).  It also doesn't seem to work very well here where we want to allow setter of `Map<Object, Object>`, which could be "populated" with values of the wrong type.

brianshannan-wf: You convinced me. I made the interface non-nats specific, and if we want to add objects to the ephemeral properties map, from a frugal server, then we can add a static getter in the server that will retrieve the object with the proper typing to aid consumers

lib/java/src/main/java/com/workiva/frugal/FContext.java
```diff
@@ -287,12 +289,12 @@ public FContext addEphemeralProperty(Object key, Object value) {
...
-     * @param ephemeralProperties ephemeral properties to add
+     * @param properties ephemeral properties to add
      * @return FContext
      */
-    public FContext addEphemeralProperties(Map<Object, Object> ephemeralProperties) {
-        for (Map.Entry<Object, Object> pair : ephemeralProperties.entrySet()) {
-            addEphemeralProperty(pair.getKey(), pair.getValue());
+    public FContext addEphemeralProperties(Map<Object, Object> properties) {
+        for (Map.Entry<Object, Object> pair : properties.entrySet()) {
+            ephemeralProperties.putAll(properties);
```
brettkail-wk: Remove the loop.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServerEventHandler.java
```diff
@@ -4,6 +4,14 @@
 
 /**
  * Provides an interface with which to handle events from an FNatsServer.
+ *
+ * An FNatsServerEventHandler should serve a distinct purpose from middleware.
+ * A set of middleware, with a processor, should be able to be used on any kind
+ * of frugal server. Conversely, this should only be used for behaviour and
+ * events for an FNatsServer not applicable to other frugal servers.
```
brettkail-wk: Remove final sentence?

Perhaps clarify that event handler is called "around" the call to a processor and middleware, so some use-cases are for collecting timings between message receive/processing, or setting/resetting MDC around the entire request processing.

lib/java/src/main/java/com/workiva/frugal/server/FServerEventHandler.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * An FServerEventHandler should serve a distinct purpose from middleware.
+ * A set of middleware, with a processor, should be able to be used on any kind
+ * of frugal server. Conversely, this should only be used for behaviour and
+ * events for an FNatsServer not applicable to other frugal servers.
+ *
+ * It is preferred to use middleware if either solution works, as middleware is
+ * more portable between different servers.
+ */
+public interface FServerEventHandler {
```
brettkail-wk: Are you thinking to implement for `FServlet` or defer to another PR?

lib/java/src/main/java/com/workiva/frugal/FContext.java
brianshannan-wf: 🤦‍♂

lib/java/src/main/java/com/workiva/frugal/server/FServerEventHandler.java
brianshannan-wf: I'm going to defer to another pr, this has been open long enough I think we need to finish it off

yuanmwang-wf:
+1

tylerrinnan-wf:
+1

yuanmwang-wf:
@Workiva/release-management-p

PR #1082, opened 2019-02-20
Infeng-7476: Isset is wrong in frugal generated dart code

yuanmwang-wf:
### Story:
Currently `isset` seems to be wrong in generated dart code. 

Implements this here:
>Brett Kail [1:58 PM]
>with my proposal, `isSetSerializationVersion() -> serializationVersion != null` and `get serializationVersion() => serializationVersion ?? -1`

For more info see discussion here: https://workiva.slack.com/archives/CEFTREXT8/p1550518577001400

#### Reviewers:
@Workiva/product2

### DIFF
--- a/compiler/generator/dartlang/generator.go
+++ b/compiler/generator/dartlang/generator.go
@@ -661,7 +661,7 @@ func (g *Generator) generateStruct(s *parser.Struct) string {
 	// Fields
 	for _, field := range s.Fields {
 		contents += g.generateCommentWithDeprecated(field.Comment, tab, field.Annotations)
-		contents += fmt.Sprintf(tab+"%s %s%s;\n",
+		contents += fmt.Sprintf(tab+"%s _%s%s;\n",
 			g.getDartTypeFromThriftType(field.Type), toFieldName(field.Name), g.generateInitValue(field))
 		contents += fmt.Sprintf(tab+"static const int %s = %d;\n", strings.ToUpper(field.Name), field.ID)
 	}
@@ -672,12 +672,6 @@ func (g *Generator) generateStruct(s *parser.Struct) string {
 	contents += fmt.Sprintf(tab+"%s() {", s.Name)
 	if len(s.Fields) > 0 {
 		contents += "\n"
-		for _, field := range s.Fields {
-			if field.Default != nil {
-				value := g.generateConstantValue(field.Type, field.Default, tab)
-				contents += fmt.Sprintf(tabtab+"this.%s = %s;\n", toFieldName(field.Name), value)
-			}
-		}
 		contents += tab + "}\n\n"
 	} else {
 		contents += "}\n\n"
@@ -732,6 +726,7 @@ func (g *Generator) generateInitValue(field *parser.Field) string {
 func (g *Generator) generateFieldMethods(s *parser.Struct) string {
 	contents := ""
 	for _, field := range s.Fields {
+		dartType := g.getDartTypeFromThriftType(field.Type)
 		titleName := strings.Title(field.Name)
 		fName := toFieldName(field.Name)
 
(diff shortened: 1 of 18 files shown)


aviary-wf:
## Security Insights

## Action Items
---

yuanmwang-wf:
Please review: @nathancoleman-wf @patkujawa-wf @brettkail-wk

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
```diff
@@ -13,34 +13,52 @@ class Album implements thrift.TBase {
...
+
   @deprecated
-  bool isSetTracks() => tracks != null;
+  bool isSetTracks() => _tracks != null;
 
   @deprecated
-  unsetTracks() => tracks = null;
+  unsetTracks() => this._tracks = null;
+
+  double get duration => this._duration;
```
brettkail-wk: Reviewing the diff relative to before the previous PR ([link](https://github.com/Workiva/frugal/compare/2.25.2...yuanmwang-wf:fix-isset)), it seems `isSetX` should return false after constructing the instance.  With the proposed encoding (null as unset, getter does defaulting), that means the field needs to be declared as `double _duration;` rather than `= 0.0`, and then this would need to be `_duration ?? 0.0` (all the `this.` seem unnecessary?)

Well, except it looks like previously the setter was tracking whether the set method was called at all, regardless of whether the argument was null?  It's unclear whether that was actually intended (the `setFieldValue` methods at least don't seem to set the isset status...).  Perhaps my proposed encoding really only makes sense for `optional` and not for default requiredness?  I don't have confidence that I understand all the issues here, sorry.

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
```diff
@@ -13,34 +13,52 @@ class Album implements thrift.TBase {
...
   static const int DURATION = 2;
-  String aSIN;
+  String _aSIN;
   static const int ASIN = 3;
 
 
   Album() {
   }
 
+  List<t_v1_music.Track> get tracks => this._tracks;
```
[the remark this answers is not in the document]
patkujawa-wf: Default values in dart (which don't necessarily map to default values provided in the frugal IDL), specifically.

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
```diff
@@ -13,34 +13,52 @@ class Album implements thrift.TBase {
   static final thrift.TField _DURATION_FIELD_DESC = new thrift.TField("duration", thrift.TType.DOUBLE, 2);
   static final thrift.TField _ASIN_FIELD_DESC = new thrift.TField("ASIN", thrift.TType.STRING, 3);
 
-  List<t_v1_music.Track> tracks;
+  List<t_v1_music.Track> _tracks;
```
patkujawa-wf: If there's no default value on the right hand side, there's no need to generate the getter and use the private field, we can just expose the field publicly.

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
```diff
@@ -13,34 +13,52 @@ class Album implements thrift.TBase {
...
   static const int TRACKS = 1;
-  double duration = 0.0;
+  double _duration = 0.0;
```
patkujawa-wf: In order for this to work, we can never have default values. For Dart, that implicitly defaults the field to `null`, which is what we need for the getter to be able to null-coalesce and provide the default value (if one is specified in the IDL).

compiler/generator/dartlang/generator.go
```diff
@@ -732,21 +726,30 @@ func (g *Generator) generateInitValue(field *parser.Field) string {
...
+		contents += fmt.Sprintf(tabtab+"this._%s = %s;\n", fName, fName)
+		contents += tab + "}\n\n"
+
 		contents += fmt.Sprintf(tab + "@deprecated\n")
-		contents += fmt.Sprintf(tab+"bool isSet%s() => %s != null%s;\n\n", titleName, fName, andIsNotDefault)
+		contents += fmt.Sprintf(tab+"bool isSet%s() => _%s != null;\n\n", titleName, fName)
 
 		contents += fmt.Sprintf(tab + "@deprecated\n")
-		contents += fmt.Sprintf(tab+"unset%s() => %s = null;\n\n", titleName, fName)
+		contents += fmt.Sprintf(tab+"unset%s() => this._%s = null;\n\n", titleName, fName)
```
patkujawa-wf: Brett is correct, there's no need for any of the `this.` prefixes because there is no scope ambiguity. That said, there's no harm in keeping it either.

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
patkujawa-wf: I believe it was intended that `isset` would return true even if consumers had set the underlying value to `null`. These changes (and the previous ones) change that semantic, but in practice I don't think that matters.

patkujawa-wf:
The example IDL (https://github.com/Workiva/frugal/blob/master/examples/idl/music.frugal) should probably include a field with a default value (e.g. `1: string title = "Unknown Title"`) as well as some fields with `required` and `optional` requiredness, so we can see & manually verify how those are being generated.

examples/dart/gen-dart/v1_music/lib/src/f_album.dart
patkujawa-wf: Looks like the logic for defaulting values is in `generateInitValue` and reflects a golang bias. The default value for any dart types (without a default value specified in the IDL) should be `null`, as even primitives are nullable objects in dart.

patkujawa-wf: Of course the risk in doing so is that we now cause null exceptions for consumers who were relying on non-null default values for generated dart code :sigh:.

patkujawa-wf:
After reviewing and thinking through, I think the safest approach is just to revert the previous changes and return to the previous semantics and defaults of `isset`. If we try to shoehorn this, we're probably going to end up with other unexpected (semantic) breaking changes.

brianshannan-wf:
I agree, we should revert the original pr

yuanmwang-wf:
Closing this, working on the revert pr now.

PR #1084, opened 2019-02-25
Infeng-7520: Update jnats to latest version 2.4.1 in frugal

yuanmwang-wf:
### Story:
Updates jnats to latest 2.4.1, basically reopens https://github.com/Workiva/frugal/pull/1035.

### Acceptance Criteria:

#### Reviewers:
@Workiva/product2

### DIFF
--- a/examples/java/src/main/java/examples/NatsClient.java
+++ b/examples/java/src/main/java/examples/NatsClient.java
@@ -15,7 +15,6 @@ import v1.music.FStore;
 
 import java.io.IOException;
 import java.util.concurrent.TimeoutException;
-import java.util.Properties;
 
 /**
  * Creates a NATS client.
--- a/examples/java/src/main/java/examples/NatsClient.java
+++ b/examples/java/src/main/java/examples/NatsClient.java
@@ -6,7 +6,7 @@ import com.workiva.frugal.provider.FServiceProvider;
 import com.workiva.frugal.transport.FNatsTransport;
 import com.workiva.frugal.transport.FTransport;
 import io.nats.client.Connection;
-import io.nats.client.ConnectionFactory;
+import io.nats.client.Options;
 import io.nats.client.Nats;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TBinaryProtocol;
@@ -15,20 +15,23 @@ import v1.music.FStore;
 
 import java.io.IOException;
 import java.util.concurrent.TimeoutException;
+import java.util.Properties;
 
 /**
  * Creates a NATS client.
  */
 public class NatsClient {
 
-    public static void main(String[] args) throws IOException, TimeoutException, TException {
+    public static void main(String[] args) throws IOException, TimeoutException, TException, InterruptedException {
         // Specify the protocol used for serializing requests.
         // The protocol stack must match the protocol stack of the server.
         FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
 
         // Create a NATS client (using default options for local dev)
-        ConnectionFactory cf = new ConnectionFactory(Nats.DEFAULT_URL);
-        Connection conn = cf.createConnection();
+        Properties properties = new Properties();
+        properties.put(Options.PROP_URL, Options.DEFAULT_URL);
+        Options.Builder optionsBuilder = new Options.Builder(properties);
+        Connection conn = Nats.connect(optionsBuilder.build());
 
         // Create and open a new transport that uses NATS for sending data.
         // The NATS transport will communicate using the music-service topic.
--- a/examples/java/src/main/java/examples/NatsClient.java
+++ b/examples/java/src/main/java/examples/NatsClient.java
@@ -14,14 +14,13 @@ import v1.music.Album;
 import v1.music.FStore;
 
 import java.io.IOException;
-import java.util.concurrent.TimeoutException;
 
 /**
  * Creates a NATS client.
  */
 public class NatsClient {
 
-    public static void main(String[] args) throws IOException, TimeoutException, TException, InterruptedException {
+    public static void main(String[] args) throws IOException, TException, InterruptedException {
         // Specify the protocol used for serializing requests.
         // The protocol stack must match the protocol stack of the server.
         FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
--- a/examples/java/src/main/java/examples/NatsClient.java
+++ b/examples/java/src/main/java/examples/NatsClient.java
@@ -28,9 +28,7 @@ public class NatsClient {
         FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
 
         // Create a NATS client (using default options for local dev)
-        Properties properties = new Properties();
-        properties.put(Options.PROP_URL, Options.DEFAULT_URL);
-        Options.Builder optionsBuilder = new Options.Builder(properties);
+        Options.Builder optionsBuilder = new Options.Builder().server(Options.DEFAULT_URL);
         Connection conn = Nats.connect(optionsBuilder.build());
 
         // Create and open a new transport that uses NATS for sending data.
--- a/examples/java/src/main/java/examples/NatsSubscriber.java
+++ b/examples/java/src/main/java/examples/NatsSubscriber.java
@@ -6,31 +6,28 @@ import com.workiva.frugal.transport.FNatsPublisherTransport;
 import com.workiva.frugal.transport.FNatsSubscriberTransport;
 import com.workiva.frugal.transport.FPublisherTransportFactory;
 import com.workiva.frugal.transport.FSubscriberTransportFactory;
+import v1.music.AlbumWinnersSubscriber;
+
 import io.nats.client.Connection;
 import io.nats.client.Nats;
 import io.nats.client.Options;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TBinaryProtocol;
-import v1.music.AlbumWinnersSubscriber;
 
 import java.io.IOException;
-import java.util.Properties;
-import java.util.concurrent.TimeoutException;
 
 /**
  * Create a NATS PubSub subscriber.
  */
 public class NatsSubscriber {
 
-    public static void main(String[] args) throws TException, IOException, TimeoutException {
+    public static void main(String[] args) throws TException, IOException, InterruptedException {
         // Specify the protocol used for serializing requests.
         // The protocol stack must match the protocol stack of the publisher.
         FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
 
         // Create a NATS client (using default options for local dev)
-        Properties properties = new Properties();
-        properties.put(Options.PROP_URL, Options.DEFAULT_URL);
-        Options.Builder optionsBuilder = new Options.Builder(properties);
+        Options.Builder optionsBuilder = new Options.Builder().server(Options.DEFAULT_URL);
         Connection conn = Nats.connect(optionsBuilder.build());
 
         // Create the pubsub scope provider, given the NATs connection and protocol
--- a/examples/java/src/main/java/examples/NatsSubscriber.java
+++ b/examples/java/src/main/java/examples/NatsSubscriber.java
@@ -7,13 +7,14 @@ import com.workiva.frugal.transport.FNatsSubscriberTransport;
 import com.workiva.frugal.transport.FPublisherTransportFactory;
 import com.workiva.frugal.transport.FSubscriberTransportFactory;
 import io.nats.client.Connection;
-import io.nats.client.ConnectionFactory;
 import io.nats.client.Nats;
+import io.nats.client.Options;
 import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TBinaryProtocol;
 import v1.music.AlbumWinnersSubscriber;
 
 import java.io.IOException;
+import java.util.Properties;
 import java.util.concurrent.TimeoutException;
 
 /**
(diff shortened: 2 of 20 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
```diff
@@ -96,42 +96,39 @@ public boolean isOpen() {
...
     }
 
     /**
      * Unsubscribes from the inbox subject and closes the response buffer.
      */
     @Override
     public void close() {
-        if (sub == null) {
+        if (dispatcher == null) {
             return;
```
kevinfrommelt-wf: Need to make sure we always call `super.close()`
```java
    @Override
    public void close() {
        if (dispatcher != null) {
            try {
                dispatcher.unsubscribe(subject);
            } catch (IllegalStateException e) {
                LOGGER.warn("NATS transport could not unsubscribe from subscription: " + e.getMessage());
            }
            dispatcher = null;
        }
        super.close();
    }
```

examples/java/src/main/java/examples/NatsSubscriber.java
```diff
@@ -27,8 +28,10 @@ public static void main(String[] args) throws TException, IOException, TimeoutEx
         FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
 
         // Create a NATS client (using default options for local dev)
-        ConnectionFactory cf = new ConnectionFactory(Nats.DEFAULT_URL);
-        Connection conn = cf.createConnection();
+        Properties properties = new Properties();
+        properties.put(Options.PROP_URL, Options.DEFAULT_URL);
+        Options.Builder optionsBuilder = new Options.Builder(properties);
```
brianshannan-wf: nit: Doesn't matter much because these are just examples, but I think it would be more clear to use the methods on the options builder instead of using an opaque properties object, like
```Option.Builder optionsBuilder = new Options.Builder().server(Options.DEFAULT_URL);```

File: lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java [hunk 3]
```diff
@@ -63,13 +62,12 @@ protected FNatsSubscriberTransport(Connection conn, String queue) {
...
         /**
-         * Creates a NATS FSubscriberTransportFactory using the provided NATS connection. Subscribers using this
-         * transport will not use a queue.
+         * Creates a NATS FSubscriberTransportFactory using the provided NATS connection.
          *
          * @param conn NATS connection
          */
         public Factory(Connection conn) {
-            this(conn, null);
+            this(conn, "");
```
brianshannan-wf: nit: Is there a reason to change this from null? I find null to be more intuitive (in java) to represent no string, rather than a present string that is empty

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -125,21 +123,28 @@ public void subscribe(String topic, FAsyncCallback callback) throws TException {
...
+        if (dispatcher == null) {
             LOGGER.warn("attempted to unsubscribe with a null internal " +
                     "subscription - possibly unsubscribing more than once - subject: " + subject);
             return;
         }
         try {
-            sub.unsubscribe();
-        } catch (IOException e) {
+            dispatcher.unsubscribe(getFormattedSubject());
+        } catch (IllegalStateException e) {
```
brianshannan-wf: Should we be catching this? I think this is a runtime exception in java, and they generally aren't caught. It looks like this is thrown when the dispatcher (or connection) is already closed (https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/NatsDispatcher.java#L201). I would lean toward not catching this, but would welcome other opinions.

lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
```diff
@@ -96,42 +96,38 @@ public boolean isOpen() {
...
-            return;
-        }
-        try {
-            sub.unsubscribe();
-        } catch (IOException e) {
-            LOGGER.warn("NATS transport could not unsubscribe from subscription: " + e.getMessage());
+        if (dispatcher != null) {
+            try {
+                dispatcher.unsubscribe(subject);
+            } catch (IllegalStateException e) {
```
brianshannan-wf: Same comment about catching this exception

test/integration/java/frugal-integration-test/src/main/java/com/workiva/TestServer.java
```diff
@@ -79,8 +80,10 @@ public static void main(String[] args) {
             TProtocolFactory protocolFactory = whichProtocolFactory(protocolType);
             FProtocolFactory fProtocolFactory = new FProtocolFactory(protocolFactory);
 
-            ConnectionFactory cf = new ConnectionFactory("nats://localhost:4222");
-            Connection conn = cf.createConnection();
+            Properties properties = new Properties();
+            properties.put(Options.PROP_URL, Options.DEFAULT_URL);
```
brianshannan-wf: Same comment about using the options builder

scripts/skynet/cross/setup_java.sh
```diff
@@ -16,7 +16,6 @@ cd ${FRUGAL_HOME}/test/integration/java/frugal-integration-test
 mvn clean install:install-file -Dfile=frugal.jar -U -q
 
 # Compile java tests
-mvn clean compile assembly:single -U -q
-
+mvn clean compile assembly:single -Dfile.encoding=UTF8 -U -q
```
brianshannan-wf: Can we put this encoding in the pom so we don't have to specify this on individual commands?

examples/java/src/main/java/examples/NatsClient.java
```diff
@@ -15,20 +15,23 @@
...
+    public static void main(String[] args) throws IOException, TimeoutException, TException, InterruptedException {
         // Specify the protocol used for serializing requests.
         // The protocol stack must match the protocol stack of the server.
         FProtocolFactory protocolFactory = new FProtocolFactory(new TBinaryProtocol.Factory());
 
         // Create a NATS client (using default options for local dev)
-        ConnectionFactory cf = new ConnectionFactory(Nats.DEFAULT_URL);
-        Connection conn = cf.createConnection();
+        Properties properties = new Properties();
+        properties.put(Options.PROP_URL, Options.DEFAULT_URL);
```
brianshannan-wf: There were a couple of other places we're still using properties, could you change those? Other than that looks good

yuanmwang-wf:
@brianshannan-wf sorry I missed those, they should be fixed now.

brianshannan-wf:
+1, would like to do a little qa on this. I'll try to do that this afternoon

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -340,11 +343,7 @@ private void process() {
             }
 
             // Send response.
-            try {
-                conn.publish(reply, output.getWriteBytes());
-            } catch (IOException e) {
-                LOGGER.warn("failed to request response: " + e.getMessage());
-            }
+            conn.publish(reply, output.getWriteBytes());
```
kevinfrommelt-wf: This can throw a `IllegalStateException`. What happens if we don't catch it? Is it worth still logging something?

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -329,7 +332,7 @@ private void process() {
             } catch (RuntimeException e) {
                 try {
                     conn.publish(reply, output.getWriteBytes());
-                    conn.flush();
+                    conn.flush(Duration.ofSeconds(60));
```
kevinfrommelt-wf: Do we have a good mechanism for making this configurable? and if so it there value in doing that?

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brianshannan-wf: `IllegalStateException` extends `RuntimeException` in java and from what I understand, generally isn't caught, as when it happens it represents using a closed resource, or something similar.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brianshannan-wf: I think making this configurable is something we can address after cutting a major version, as we don't provide that now. I think ideally this would reflect the timeout in the FContext

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
yuanmwang-wf: I guess it's the same for `IllegalArgumentException` too? 

https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/NatsConnection.java#L627

examples/java/src/main/java/examples/NatsClient.java
```diff
@@ -21,14 +21,14 @@
  */
 public class NatsClient {
 
-    public static void main(String[] args) throws IOException, TimeoutException, TException {
+    public static void main(String[] args) throws IOException, TimeoutException, TException, InterruptedException {
```
kevinfrommelt-wf: Is `TimeoutException` needed here? Seems like you removed it in all the other examples

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
yuanmwang-wf: Yeah I only did this because in jnats 1 it was hardcoded to be 1 minute, and the function changed to take a timeout in jnats 2 now. 

In jnats 1: https://github.com/nats-io/java-nats/blob/71f4154ef7a3037356f79a3f63fc20602b505c45/src/main/java/io/nats/client/ConnectionImpl.java#L1757

examples/java/src/main/java/examples/NatsClient.java
yuanmwang-wf: You're right, it doesn't need to be here.

lib/java/src/main/java/com/workiva/frugal/transport/FNatsPublisherTransport.java
```diff
@@ -109,12 +107,7 @@ public void publish(String topic, byte[] payload) throws TTransportException {
                     String.format("Message exceeds %d bytes, was %d bytes",
                             NATS_MAX_MESSAGE_SIZE, payload.length));
         }
-
-        try {
-            conn.publish(getFormattedSubject(topic), payload);
-        } catch (IOException e) {
-            throw new TTransportException("publish: unable to publish data: " + e.getMessage());
-        }
+        conn.publish(getFormattedSubject(topic), payload);
```
kevinfrommelt-wf: This can throw a IllegalStateException. What happens if we don't catch it? Is it worth still logging something?

brianshannan-wf:
+1

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -125,21 +123,24 @@ public void subscribe(String topic, FAsyncCallback callback) throws TException {
...
                     "subscription - possibly unsubscribing more than once - subject: " + subject);
             return;
         }
-        try {
-            sub.unsubscribe();
-        } catch (IOException e) {
-            LOGGER.warn("could not unsubscribe from subscription. " + e.getMessage());
-        }
-        sub = null;
+        dispatcher.unsubscribe(getFormattedSubject());
```
kevinfrommelt-wf: This can throw `IllegalStateException` and `IllegalArgumentException`. Do we need to handle them?

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -125,21 +123,24 @@ public void subscribe(String topic, FAsyncCallback callback) throws TException {
...
+        conn.closeDispatcher(dispatcher);
```
kevinfrommelt-wf: This can throw IllegalStateException and IllegalArgumentException. Do we need to handle them?

lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
```diff
@@ -86,7 +86,7 @@ public FNatsTransport withInbox(String inbox) {
      */
     @Override
     public boolean isOpen() {
-        return sub != null && conn.getState() == Nats.ConnState.CONNECTED;
+        return dispatcher != null && conn.getStatus() == Status.CONNECTED;
```
kevinfrommelt-wf: Should we check `dispatcher.isActive()`

kevinfrommelt-wf: Maybe not since it didn't do that before

kevinfrommelt-wf:
+1

lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
```diff
@@ -152,16 +144,23 @@ public void onMessage(Message message) {
...
         }
         return new TTransportException(TTransportExceptionType.NOT_OPEN,
                 String.format("%s NATS Transport not open", prefix));
     }
+
+    /**
+     * Helper to generate a random id for inbox.
+     */
+    private static String createInbox() {
+        return INBOX_PREFIX + UUID.randomUUID().toString().replace("-", "");
```
kevinfrommelt-wf: Replace `INBOX_PREFIX` with `conn.getOptions().getInboxPrefix()`

kevinfrommelt-wf:
+1

brianshannan-wf:
+10

kevinfrommelt-wf:
@Workiva/release-management-p

PR #1097, opened 2019-03-29
Allow setting the stop timeout for stopping a FNatsServer

brandonbodnar-wk:
### Story:
When stopping a FNatsServer, the stop method awaits the server's
configured executor service for up to a defined number of seconds.
This is to give the server time to finish processing current tasks.

This time use to be hard coded to 30 seconds. Different servers may
have different needs, however. Allowing callers to both configure
the stop timeout for the server or to specify a timeout while calling
stop will enable additional use cases where a timeout other than the
default is desired.

### Acceptance Criteria:

### Design Notes:
Existing behavior should not change.

### How To Test:
Unit testing added.

### My Test Results:
Passed locally.

#### Reviewers:
@Workiva/product2

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -45,6 +45,7 @@ public class FNatsServer implements FServer {
     private static final Logger LOGGER = LoggerFactory.getLogger(FNatsServer.class);
     public static final int DEFAULT_WORK_QUEUE_LEN = 64;
     public static final int DEFAULT_WATERMARK = 5000;
+    public static final int DEFAULT_STOP_TIMEOUT = 30;
 
     private final Connection conn;
     private final FProcessor processor;
@@ -53,6 +54,7 @@ public class FNatsServer implements FServer {
     private final String[] subjects;
     private final String queue;
     private final long highWatermark;
+    private final long stopTimeout;
 
     private final CountDownLatch shutdownSignal = new CountDownLatch(1);
     private final ExecutorService executorService;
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

yuanmwang-wf:
Closing/reopening so workiva build can pick up the new base branch.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -247,10 +265,22 @@ public void serve() throws TException {
...
+
+    /**
+     * Stops the server by shutting down the executor service processing tasks. Attempts to await
+     * existing tasks completing by up to the given timeout.
+     *
+     * @param timeout the maximum time to wait
+     * @param unit the time unit of the timeout argument
+     * @throws TException if the server fails to stop
+     */
+    public void stop(long timeout, TimeUnit unit) throws TException {
```
yuanmwang-wf: Does this need to be public?

brianshannan-wf: Yeah, that's the new method they want to call

brianshannan-wf: Actually that's a good point, I don't think this needs to be public if the stop timeout is set in the builder. Having both seems redundant, I'm not sure which would be preferred though

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -189,6 +194,18 @@ public Builder withHighWatermark(long highWatermark) {
...
         }
 
+        /**
+         * Controls how many seconds to attempt wait for existing tasks to complete when stopping
+         * the server via {@link FNatsServer#stop()}.
+         *
+         * @param stopTimeout duration in seconds
+         * @return Builder
+         */
+        public Builder withStopTimeout(long stopTimeout) {
```
brettkail-wk: Alternatively, `withStopTimeout(long stopTimeout, TimeUnit unit)` with `this.stopTimeoutNanos = unit.toNanos(stopTimeout);` might be less error-prone for callers.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: I'm good with not making this public (actually, will probably just remove it). For our use case, we will only use the Builder. 

I had added this in case someone wanted to call stop with a dynamic timeout value. But we can wait until someone actually wants that and add it back later.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: Good point. I had thought of this but did not do it because another method just takes milliseconds instead of a TimeUnit. But I prefer the TimeUnit method too, so I will change to that.

brettkail-wk: Yeah, having one method accept unqualified ms and another accepting unqualified sec seemed like it would just cause confusion :-).

yuanmwang-wf:
+1

kevinfrommelt-wf:
+1

PR #1098, opened 2019-04-02
Wait for unsubscribes to finish before shutting down executor service.

brandonbodnar-wk:
### Story:
Previously, if a stop signal came in, the executor service was shutdown
prior to unsubscribing to the nats messages for the server. This would
lead to nats messages still being delivered to the server despite that
the messages will never be executed / acted upon because the executor
service was shutdown.

This changes the stop / shutdown process to first wait for the
unsubscriptions to finish prior to shutting down the executor service.
Thus, no new messages should be received from the nats server while
this frugal server is shutting down.

### Acceptance Criteria:

### Design Notes:
While it may be intended that people only call `serve` once and `stop` once on a FServer instance, I tried to design this so that if someone had a bug where they were calling these methods multiple times that this change should still work (blocking and not blocking the calls as before). An example of this is two serve calls executing in parallel where one of the threads is interrupted but not the other. As before, this change unsubscribes the interrupted server and closes its dispatcher before exiting the serve method. The other parallel serve method should remain unaffected.

### How To Test:
Added some unit tests to cover the existing behavior.

### My Test Results:
Unit testing passed locally.

#### Reviewers:
@Workiva/product2

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -273,14 +273,8 @@ public class FNatsServer implements FServer {
             }
             LOGGER.info("Frugal server stopping...");
 
-            for (String subject : subjects) {
-                try {
-                    dispatcher.unsubscribe(subject);
-                } catch (IllegalStateException e) {
-                    LOGGER.warn("Frugal server failed to unsubscribe from " + subject + ": " + e
-                          .getMessage());
-                }
-            }
+            // Closing the dispatcher will unsubscribe from all current subscriptions for that
+            // dispatcher
             conn.closeDispatcher(dispatcher);
             // If serving stopped due to stop being called, we want to give the signal the stop method
             // that this serve has completed its actions and the stop method can move on to the next
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -245,6 +245,8 @@ public class FNatsServer implements FServer {
             boolean shutdownSignalReceived;
             try {
                 shutdownSignal.await();
+                // This shutdown signal may have been received from a prior call to stop if
+                // serve was inadvertantly called on an already stopped server.
                 shutdownSignalReceived = true;
             } catch (InterruptedException ignored) {
                 shutdownSignalReceived = false;
@@ -267,8 +269,19 @@ public class FNatsServer implements FServer {
             // may not have been interrupted and we don't want to be blocked by them.
             if (shutdownSignalReceived) {
                 // Advance to next state now that all unsubscribes are finished
+                // A phaser is used here because while we expect callers to have only issued a
+                // single call to serve we do not currently enforce that. If someone is running
+                // serve on the same instance in different threads, then we want to ensure all
+                // unsubscribes are finished before moving on to shutting down. The phaser allows
+                // us to notify when each call to serve is ready to move on to the next phase.
+                // While we could add in some type of enforcement for not allowing more than one
+                // call to serve or a call to serve after stop, that would likely be a breaking
+                // change to callers doing this now.
                 partiesAwaitingFullShutdown.arriveAndAwaitAdvance();
-                // Wait for full shutdown to finish
+                // Wait for full shutdown to finish (if this was a call to serve after stop was
+                // already called, then the shutdown would already be finished. But since stop
+                // never registered with the phaser, then this below method should immediately
+                // return as no other parties would be waiting.)
                 partiesAwaitingFullShutdown.arriveAndAwaitAdvance();
             }
         } finally {
@@ -288,7 +301,9 @@ public class FNatsServer implements FServer {
             // Unblock serving threads to allow them to unsubscribe
             shutdownSignal.countDown();
 
-            // Wait for all unsubscriptions to finish
+            // Wait for all unsubscriptions to finish if serve has been called and is currently
+            // serving. If serve has not been called or all calls to serve have already exited,
+            // then this call should be non-blocking.
             partiesAwaitingFullShutdown.arriveAndAwaitAdvance();
 
             // Attempt to perform an orderly shutdown of the worker pool by trying to complete any in-flight requests.
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -56,6 +56,20 @@ public class FNatsServer implements FServer {
     private final long highWatermark;
 
     private final CountDownLatch shutdownSignal = new CountDownLatch(1);
+    /**
+     * Synchronizes shutdown between {@link #serve} and {@link #stop} threads.
+     * The implied states are:
+     * <ol>
+     * <li>Serve - initial state; serve thread waits, and stop thread calls
+     * {@link CountDownLatch#countDown()} on {@link FNatsServer#shutdownSignal} before
+     * advancing to the next state
+     * <li>Unsubscribe - stop thread waits, and serve thread unsubscribes before
+     * advancing to the next state
+     * <li>Shutdown executor service - serve thread waits, and stop thread awaits
+     * executor service shutdown before advancing to the next state
+     * <li>Shutdown - serve and stop threads both complete
+     * </ol>
+     */
     private final Phaser partiesAwaitingFullShutdown = new Phaser();
     private final ExecutorService executorService;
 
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -224,20 +226,39 @@ public void serve() throws TException {
...
-        for (String subject : subjects) {
+            boolean shutdownSignalReceived = false;
             try {
-                dispatcher.unsubscribe(subject);
-            } catch (IllegalStateException e) {
-                LOGGER.warn("Frugal server failed to unsubscribe from " + subject + ": " + e.getMessage());
+                shutdownSignal.await();
+                shutdownSignalReceived = true;
+            } catch (InterruptedException ignored) {
+            }
```
brettkail-wk: IMO clearer with final var semantics:
```
boolean shutdownSignalReceived;
try {
  ...
  shutdownSignalReceived = true;
} catch (InterruptedException e) {
  shutdownSignalReceived = false;
}
```

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -224,20 +226,39 @@ public void serve() throws TException {
...
+                }
             }
+            conn.closeDispatcher(dispatcher);
+            // If serving stopped due to stop being called, we want to give the signal the stop method
+            // that this serve has completed its actions and the stop method can move on to the next
+            // phase of the shutdown. If shutdown signal was not received (i.e., serve is exiting
+            // because the thread was interrupted, then we do not want to wait as other calls to serve
+            // may not have been interrupted and we don't want to be blocked by them.
+            if (shutdownSignalReceived) {
+                // Wait for all unsubscribes to finish
```
brettkail-wk: IMO clearer as "Advance to next state now that all unsubscribes are complete".

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -224,20 +226,39 @@ public void serve() throws TException {
...
+                partiesAwaitingFullShutdown.arriveAndAwaitAdvance();
```
brettkail-wk: If `dispatcher.unsubscribe` or `closeDispatcher` throw, then this is never reached, and `stop` hangs indefinitely?  It appears some additional `try`/`finally` would be needed.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -55,6 +56,7 @@
     private final long highWatermark;
 
     private final CountDownLatch shutdownSignal = new CountDownLatch(1);
```
brettkail-wk: > While it may be intended that people only call serve once and stop once on a FServer instance, I tried to design this so that if someone had a bug where they were calling these methods multiple times that this change should still work (blocking and not blocking the calls as before). 

Can instances be meaningfully reused given that `shutdownSignal` is not currently recreated?  I expect the second call to `serve` to instantly unsubscribe.  If my understanding is correct, can the design of this PR be simplified?

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -224,20 +226,39 @@ public void serve() throws TException {
...
-                LOGGER.warn("Frugal server failed to unsubscribe from " + subject + ": " + e.getMessage());
+                shutdownSignal.await();
+                shutdownSignalReceived = true;
+            } catch (InterruptedException ignored) {
+            }
+            LOGGER.info("Frugal server stopping...");
+
+            for (String subject : subjects) {
+                try {
+                    dispatcher.unsubscribe(subject);
```
brettkail-wk: @brianshannan-wf Pre-existing, do you know why `unsubscribe` is called manually rather than just relying on `conn.closeDispatcher` to do that?

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -55,6 +56,7 @@
     private final long highWatermark;
 
     private final CountDownLatch shutdownSignal = new CountDownLatch(1);
+    private final Phaser partiesAwaitingFullShutdown = new Phaser();
```
brettkail-wk: I have yet to find a satisfying and simple multi-threaded state machine primitive in Java.  I've used `CyclicBarrier` but not `Phaser`, so I had to read up; it appears to be about the same but more flexible.  My problem with both is that it's easy to create state machines where the states are implicit and spread across different code.  Perhaps add something like this as javadoc to make the states explicit?
```
/**
 * Synchronizes shutdown between {@link #serve} and {@link #stop} threads.
 * The implied states are:
 * <ol>
 * <li>Serve - initial state; serve thread waits, and stop thread calls
 * {@link shutdownSignal#countDown} before advancing to the next state
 * <li>Unsubscribe - stop thread waits, and serve thread unsubscribes before
 * advancing to the next state
 * <li>Shutdown executor service - serve thread waits, and stop thread awaits
 * executor service shutdown before advancing to the next state
 * <li>Shutdown - serve and stop threads both complete
 * </ol>
 */
```

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: That should be handled by the finally on line 260 I think. If there is an error in closing the dispatcher or otherwise an uncaught exception, the `partiesAwaitingFullShutdown.arriveAndDeregister();` should mark this party as arrived and deregister its interest for any future phases. That should be sufficient to ensure the stop is not waiting for the serve thread to finish.

We could potentially do a `awaitAdvanceInterruptibly` with a timeout in the stop thread also if we want extra safety that it will not be blocked forever.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: That is a correct understanding and one that I added a unit test to verify that it was the correct existing behavior. If someone calls serve after stop, it immediately subs then unsubs and returns because the latch is already at 0.

However, my concern was someone with an existing bug that either does this (calls serve after a stop) or that calls serve multiple times on the same server before calling stop. I wanted to make sure that this change does not result in their code from going from a (1) buggy but immediately exits state to a (2) hangs forever state.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: I had this same question. I assumed it was for the extra logging in the catch, but that was just an assumption.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: I like this documentation idea.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brettkail-wk: I overlooked this, thanks.  Looks fine.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brettkail-wk: Perhaps drop a quick comment in the code and/or test that summarizes that second paragraph to explain why the code behaves the way it does.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
yuanmwang-wf: This probably happened because dispatcher was not a thing in jnats 1 so in frugal we have to unsub each one manually, when moving to jnats 2 I just ported the code over. I think it's safe to call `closeDispatcher` here without manually unsubscribe everyone.

brianshannan-wf:
@brandonbodnar-wk can you clean up the merge conflicts?

brandonbodnar-wk:
@brianshannan-wf, merge conflicts should be resolved now. Thanks.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brandonbodnar-wk: Happy to remove the extra unsubscribe calls. Will do that now.

brianshannan-wf:
+1

yuanmwang-wf:
@Workiva/release-management-p

PR #1164, opened 2019-12-17
INFENG-10230 use deep equals for dart collections

brianshannan-wf:
Dart collections (lists, sets, maps) don't implement and equals method by default. The standard seems to be to use `package:collection/collection.dart` to do this, if that's not the proper way to do things in dart it should be easy to change it, but this appears to work.

For some reason `yum autoremove` decided it needed to remove systemd which didn't go well so I got rid of the autoremove command.

@Workiva/product2 @sagesmith-wf


aviary-wf:
## Security Insights

## Action Items
---

kevinfrommelt-wf:
+1

compiler/generator/dartlang/generator.go
```diff
@@ -1243,6 +1244,15 @@ func (g *Generator) generateEquals(s *parser.Struct) string {
 	return contents
 }
 
+func (g *Generator) generateFieldEquals(field *parser.Field) string {
+	fieldName := toFieldName(field.Name)
+	if !g.isDartCollection(field.Type) {
+		return fmt.Sprintf("this.%s == o.%s", fieldName, fieldName)
+	}
+
+	return fmt.Sprintf("DeepCollectionEquality().equals(this.%s, o.%s)", fieldName, fieldName)
```
brettkail-wk: I guess `SetEquality`, `ListEquality`, and `MapEquality` could be used rather than using `DeepCollectioneQuality`, which has to do runtime checks to figure out which one to use?

brianshannan-wf: The issue with those is they didn't appear to handle nested lists/sets/maps. We could use those in the generated code I think, but constructing the appropriate method call would be too complicated to be worth it, imo

brettkail-wk: i.e., `list<list<...>>` etc., where the list/set/map is a direct child of another list/set/map, so we would have to detect those cases specifically?  Then what you have makes sense, thanks.

brianshannan-wf:
@Workiva/release-management-p

PR #1170, opened 2020-01-07
Add JSON generator

brettkail-wk:
### Story:
We would like to be able to introspect Frugal files without the complexity of actually parsing the Frugal file format.

### Acceptance Criteria:

### Design Notes:
A new `-gen json` output mode is added that produces a lightweight JSON format.  The key names are deliberately kept short to minimize the size of the file since we would like to use the output in very space-constrained storage where compression is not an option.

### How To Test:
Unit test added for `variety.frugal`.

#### Reviewers:
@Workiva/product2

### DIFF
--- a/compiler/generator/json/generator.go
+++ b/compiler/generator/json/generator.go
@@ -124,9 +124,6 @@ type scope struct {
 	Operations map[string]*frugalType `json:"o,omitempty"`
 }
 
-type operation struct {
-}
-
 type method struct {
 	Params  map[int]*field `json:"p,omitempty"`
 	Results map[int]*field `json:"r,omitempty"`
--- a/compiler/generator/json/generator.go
+++ b/compiler/generator/json/generator.go
@@ -31,7 +31,7 @@ type Generator struct {
 	generated bool
 }
 
-var gen *Generator
+var _ generator.ProgramGenerator = (*Generator)(nil)
 
 // NewGenerator returns a generator for JSON descriptor files.
 func NewGenerator(options map[string]string) generator.ProgramGenerator {
--- a/compiler/generator/json/generator.go
+++ b/compiler/generator/json/generator.go
@@ -1,0 +1,302 @@
+/*
+ * Copyright 2020 Workiva
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package json
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/Workiva/frugal/compiler/generator"
+	"github.com/Workiva/frugal/compiler/parser"
+)
+
+const (
+	defaultOutputDir = "gen-json"
+)
+
+// Generator generates JSON descriptor files.
+type Generator struct {
+	options   map[string]string
+	generated bool
+}
+
+var gen *Generator
+
+// NewGenerator returns a generator for JSON descriptor files.
+func NewGenerator(options map[string]string) generator.ProgramGenerator {
+	return &Generator{
+		options: options,
+	}
+}
+
+// GetOutputDir returns the dir unchanged.
+func (*Generator) GetOutputDir(dir string, frugal *parser.Frugal) string {
+	return dir
+}
+
+// DefaultOutputDir returns "gen-json".
+func (*Generator) DefaultOutputDir() string {
+	return defaultOutputDir
+}
+
+// UseVendor returns false.
+func (*Generator) UseVendor() bool {
+	return false
+}
+
+// Generate writes a type library to the output directory.
+func (g *Generator) Generate(pf *parser.Frugal, outputDir string) error {
+	if g.generated {
+		return nil
+	}
+	g.generated = true
+
+	f, err := os.OpenFile(outputDir+"/frugal.json", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	frugals := collectFrugals(pf, []*parser.Frugal{}, map[string]struct{}{})
+	ff := toFiles(frugals)
+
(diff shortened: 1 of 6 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

compiler/generator/json/generator.go
```diff
@@ -0,0 +1,302 @@
...
+	return t
+}
+
+func toRawType(pt *parser.Type) *frugalType {
+	if pt.IsContainer() {
+		switch {
+		case pt.Name == "list":
+			return &frugalType{Value: toType(pt.ValueType)}
+		case pt.Name == "set":
+			return &frugalType{Key: toType(pt.ValueType)}
```
yuanmwang-wf: Should this return `&frugalType{Key: toType(pt.KeyType)}` instead of `pt.ValueType`?

brettkail-wk: The parser populates [pt.ValueType](https://github.com/Workiva/frugal/blob/5cff5467fcb0718c616cfb4a6bf580118903f82a/compiler/parser/grammar.peg#L404-L407) for sets, so it's correct as-is.  The JSON deliberately deviates from this internal model and looks like this:
* `map<i32,string>` - `{"k":{"b":"i32"},"v":{"b":"string"}}` (key+value)
* `set<string>` - `{"k": {"b":"string"}}` (just key, ala Go `map[string]struct{}`)
* `list<string>` - `{"v": {"b":"string"}}` (just value)

That encoding scheme seemed reasoanbly elegant to me.  I considered `{"v":{"b":"string"},"s":true}`, but that's slightly larger without any real benefit.

yuanmwang-wf: That makes sense, thanks for the clarification.

compiler/generator/json/generator.go
```diff
@@ -0,0 +1,302 @@
...
+	defaultOutputDir = "gen-json"
+)
+
+// Generator generates JSON descriptor files.
+type Generator struct {
+	options   map[string]string
+	generated bool
+}
+
+var gen *Generator
```
brianshannan-wf: Did you mean for this to be an assertion this struct satisfies the `ProgramGenerator` interface?

compiler/generator/json/generator.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// UseVendor returns false.
+func (*Generator) UseVendor() bool {
+	return false
+}
+
+// Generate writes a type library to the output directory.
+func (g *Generator) Generate(pf *parser.Frugal, outputDir string) error {
+	if g.generated {
```
brianshannan-wf: Is this boolean meant to prevent generating any code for included frugal files if the recursive flag is provided?

compiler/generator/json/generator.go
brettkail-wk: Oops, I did multiple iterations of this: assertion for `LanguageGenerator`, single global variable hack, and then I failed to update this back to an assertion for `ProgramGenerator` when I switched to that approach.  I'll fix it.

compiler/generator/json/generator.go
brettkail-wk: Yes, but I had misread the code, so I didn't realize it required opting in to recursive.  If you think it's best, I can remove it.

As an aside, it seems somewhat weird to me the `!generate` check isn't at the very top of the function (or remove the param and check `Recurse` in the caller?), which means `CompiledFiles`, `MkdirAll`, and `logv` all happen for the first level of includes?

compiler/generator/json/generator.go
```diff
@@ -0,0 +1,302 @@
...
+		t.Annotations = map[string]string{}
+		for _, pa := range pt.Annotations {
+			t.Annotations[pa.Name] = pa.Value
+		}
+	}
+	return t
+}
+
+func toRawType(pt *parser.Type) *frugalType {
+	if pt.IsContainer() {
```
brianshannan-wf: Does [UnderlyingType](https://github.com/Workiva/frugal/blob/master/compiler/parser/types.go#L820) need to be used at some point in this method? I don't remember all the rules for typedefs and their names off the top of my head, so it might be unnecessary

compiler/generator/json/generator.go
```diff
@@ -0,0 +1,302 @@
...
+		case pt.Name == "map":
+			return &frugalType{
+				Key:   toType(pt.KeyType),
+				Value: toType(pt.ValueType),
+			}
+		}
+		panic("unknown container type " + pt.Name)
+	}
+
+	if _, ok := baseTypes[pt.Name]; ok {
```
brianshannan-wf: I don't think this if statement currently does anything. Either way `&frugalType{Name: pt.Name}` is returned

compiler/generator/json/generator.go
brianshannan-wf: I don't want to be too over the top, would it be possible to add a readme that documents the expected outputs for each frugal entity and what the single letter keys mean? That way consumers won't have to read through the generator code to figure out what the json output means

compiler/generator/json/generator.go
brettkail-wk: This `if` statement is populating `Base`, not `Name` (both from `pt.Name`).

brianshannan-wf: Ah yep 🤦‍♂ . If it's desired for typedefs for the base types to work like base types, they'll definitely need an `UnderlyingType()` check. It's not immediately clear to me how you want that to behave

compiler/generator/json/generator.go
brettkail-wk: The typedef names are deliberately exposed.  For example:
1. [`typedef i32 int`](https://github.com/Workiva/frugal/blob/5cff5467fcb0718c616cfb4a6bf580118903f82a/test/idl/variety.frugal#L13)
2. [`3: int someotherthing`](https://github.com/Workiva/frugal/blob/5cff5467fcb0718c616cfb4a6bf580118903f82a/test/idl/variety.frugal#L82)
3. [expected output](https://github.com/Workiva/frugal/pull/1170/files#diff-49284122a9a630e387568148caee41d2R737-R742)
```
          "3": {
            "n": "someotherthing",
            "t": {
              "n": "int"
            }
          },
```

Processors can know to process recursively (and render as much/little detail as desired).

brianshannan-wf: Okay, that desired behaviour makes sense

compiler/generator/json/generator.go
brettkail-wk: I've added a small `documentation/json.md` file ([rendered](https://github.com/brettkail-wk/frugal/blob/wip_json_generator/documentation/json.md)).

yuanmwang-wf:
@Workiva/release-management-p

PR #1245, opened 2020-04-06
INFENG-11229: Pass timeout to aiohttp ClientSession

yuanmwang-wf:
### Story:
See https://github.com/aio-libs/aiohttp/issues/2865
Fixed here: https://github.com/aio-libs/aiohttp/pull/2972

`ClientSession` from aiohttp accepts a timeout in seconds, if not set, will default to 5 minutes. We should pass timeout from FContext to `ClientSession` too. 

#### Reviewers:
@Workiva/product2

### DIFF
--- a/lib/python/frugal/aio/transport/http_transport.py
+++ b/lib/python/frugal/aio/transport/http_transport.py
@@ -13,7 +13,7 @@ import asyncio
 import base64
 
 import async_timeout
-from aiohttp.client import ClientSession
+from aiohttp.client import ClientTimeout, ClientSession
 from thrift.transport.TTransport import TTransportBase
 from thrift.transport.TTransport import TMemoryBuffer
 from thrift.transport.TTransport import TTransportException
@@ -130,12 +130,13 @@ class FHttpTransport(FTransportBase):
         # apply the default headers so their values cannot be modified
         request_headers.update(self._headers)
 
-        async with ClientSession() as session:
+        timeout = context.timeout / 1000
+        async with ClientSession(timeout=ClientTimeout(total=timeout)) as s:
             try:
-                with async_timeout.timeout(context.timeout / 1000):
-                    async with session.post(self._url,
-                                            data=payload,
-                                            headers=request_headers) \
+                with async_timeout.timeout(timeout):
+                    async with s.post(self._url,
+                                      data=payload,
+                                      headers=request_headers) \
                             as response:
                         return response.status, await response.content.read()
             except asyncio.TimeoutError:
--- a/lib/python/frugal/aio/transport/http_transport.py
+++ b/lib/python/frugal/aio/transport/http_transport.py
@@ -9,11 +9,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import asyncio
 import base64
 
-import async_timeout
-from aiohttp.client import ClientTimeout, ClientSession
+from aiohttp.client import ClientTimeout, ClientSession, ServerTimeoutError
 from thrift.transport.TTransport import TTransportBase
 from thrift.transport.TTransport import TMemoryBuffer
 from thrift.transport.TTransport import TTransportException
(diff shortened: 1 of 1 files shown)


yuanmwang-wf:
@ryanjackson-wf @jasonaguilon-wf

aviary2-wf:
## Security Insights
<details>
</details>

## Action Items
---

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -130,12 +130,13 @@ def is_open(self):
...
 
-        async with ClientSession() as session:
+        timeout = context.timeout / 1000
+        async with ClientSession(timeout=ClientTimeout(total=timeout)) as s:
             try:
-                with async_timeout.timeout(context.timeout / 1000):
-                    async with session.post(self._url,
-                                            data=payload,
-                                            headers=request_headers) \
+                with async_timeout.timeout(timeout):
```
ryanjackson-wf, line 136: Perhaps async_timeout performs other actions I am unaware of but I wonder if it is still necessary given the timeout imposed by aiohttp.

jasonaguilon-wf: It does look like `ClientSession.post` will use the session's timeout now instead of the default of 5 minutes -- which appears to now cover the aggregate time to connect and read.

yuanmwang-wf: I also wasn't sure what the behavior was in early versions before https://github.com/aio-libs/aiohttp/pull/2972 was released. Maybe for `ClientSession` we could use 5 minutes or timeout in Fcontext, whichever is longer?

ryanjackson-wf: The default timeout prior to the fix was still five minutes. The session class at that time allowed one to specify separate read and connection timeouts. I'm not sure how those were used. The fix reduced them to a single timeout value.

yuanmwang-wf: So if we use the whichever is longer between 5 minutes and timeout in Fcontext for ClientSession, it would keep the previous behavior while allowing requests with longer timeouts to go through, thoughts on that?

ryanjackson-wf: Given that this affects all Python frugal users what you've suggested sounds safest to minimize any unintended side-effects. If the context timeout is less than the default (`aiohttp.DEFAULT_TIMEOUT`) then use the default. If greater than or equal to the default, then I suggest we increment the timeout used for `ClientSession` to prevent it from short-circuiting the one here.

Alternately this `timeout` call could possibly be removed and the timeout of the `ClientSession` be used to serve both purposes. In that case it should probably use the value from the context.

yuanmwang-wf: I agree, I've updated the code to pass timeout from FContext directly to `ClientSession` and removed the custom `async_timeout.timeout`.

aiohttp [raises `ServerTimeoutErrer` on timeouts](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L442-L454) so I updated frugal to catch that and raise its own timeout error instead.

lib/python/frugal/aio/transport/http_transport.py
```diff
@@ -130,15 +128,15 @@ def is_open(self):
         # apply the default headers so their values cannot be modified
         request_headers.update(self._headers)
 
-        async with ClientSession() as session:
+        timeout = ClientTimeout(total=context.timeout / 1000)
```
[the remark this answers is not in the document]
ryanjackson-wf: It looks like `ClientTimeout.total` is of type float so division by 1000.0 might get you what you want.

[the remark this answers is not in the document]
yuanmwang-wf: I think aio is python3 only and this shouldn't be a problem?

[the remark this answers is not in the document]
ryanjackson-wf: Yes, you are right. I've done mostly 2.x and forgot that 3.x changed from "classic" to "true" division.

yuanmwang-wf:
@Workiva/release-management-p

PR #1228, opened 2020-03-19
CPLAT-10183: Add factories for creating dart frugal objects

corwinsheahan-wf:
### Story:
As part of a new client side messaging api proposed in [RFD 274](https://sandbox.wdesk.com/a/QWNjb3VudB82NzE5NTMwMjcyOTQ4MjI0/doc/4a01314bdff4427ebe7d639867a12b8c/r/-1/v/1/sec/4a01314bdff4427ebe7d639867a12b8c_28), consumers supply a factory to the client creation methods to allow the messaging-sdk to construct an instance of their frugal class, obviating the need for some extra boilerplate. 

This PR adds this factory to the dart frugal generation, so consumers get the factory boilerplate for free.

### Acceptance Criteria:
- [ ] At least one InfRe Squad 2 member has reviewed and +1'd
- [ ] Code has been tested and results documented
- [ ] Unit tests have been updated
- [ ] Updates to documentation if necessary
- [ ] Verify and document changes to any other Messaging components
- [ ] Pull request made against the 'develop' branch, not master

### Design Notes:
TODO

### How To Test:
TODO

### My Test Results:
TODO

#### Reviewers:
@Workiva/product2

PR #1296, opened 2020-06-03
java: Expose request for extensible processing

brendanbartels-wf:
### Story:
Clients and servers are inherently decoupled with NATS messaging, which presents a unique opportunity for cooperative request handling by multiple server instances. However, to allow strategies that leverage this characteristic, like [RFD 0278](https://sandbox.wdesk.com/a/QWNjb3VudB82NzE5NTMwMjcyOTQ4MjI0/doc/0302cab5f422408196f0b5bef095770d/r/-1/v/1/sec/0302cab5f422408196f0b5bef095770d_1), the POJO used to represent the request in `FNatsServer` needs to be exposed.

### Acceptance Criteria:

### Design Notes:
This PR _only_ makes the necessary code changes to expose the minimum `FNatsServer.Request` POJO. All current behavior should be preserved. The implementation of the aforementioned RFD would have to be done by consumers.

I considered keeping the POJO a `static class`, but it required a larger refactor in order to expose a way to populate the processing instance variables (`FProtocolFactory`,`FProcessor`, etc). It was ultimately much simpler to forgo copying references of these variables into the POJO, and simply use the original references on the `FNatsServer` class in the `processRequest()` method.

### How To Test:
I've copied this file `FNatsServer` into our PR that implements the design in [RFD 0278](https://sandbox.wdesk.com/a/QWNjb3VudB82NzE5NTMwMjcyOTQ4MjI0/doc/0302cab5f422408196f0b5bef095770d/r/-1/v/1/sec/0302cab5f422408196f0b5bef095770d_1) and deployed it to our development environment. Links are in the RFD.

### My Test Results:
- [x] The original request handling behavior is unchanged.
- [x] The new ability to interact with the exposed Request POJO also works as intended.

#### Reviewers:
@Workiva/product2
@natewoods-wf
@chadknight-wf
@richklein-wf

### Review Note
It is easier to review the 2 commits separately. The 2nd commit is only a formatting change, which obfuscates the actual changes a bit.

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
@@ -422,40 +422,39 @@ public class FNatsServer implements FServer {
      * @param request the request
      */
     private void processRequest(Request request) {
-            eventHandler.onRequestStarted(request.ephemeralProperties);
+        eventHandler.onRequestStarted(request.ephemeralProperties);
+
+        try {
+            // Read and process frame (exclude first 4 bytes which represent frame size).
+            TTransport input = new TMemoryInputTransport(request.frameBytes, 4, request.frameBytes.length);
+            TMemoryOutputBuffer output = new TMemoryOutputBuffer(NATS_MAX_MESSAGE_SIZE);
 
             try {
-                // Read and process frame (exclude first 4 bytes which represent frame size).
-                TTransport input = new TMemoryInputTransport(request.frameBytes, 4, request.frameBytes.length);
-                TMemoryOutputBuffer output = new TMemoryOutputBuffer(NATS_MAX_MESSAGE_SIZE);
-
+                FProtocol inputProto = inputProtoFactory.getProtocol(input);
+                inputProto.setEphemeralProperties(request.ephemeralProperties);
+                FProtocol outputProto = outputProtoFactory.getProtocol(output);
+                processor.process(inputProto, outputProto);
+            } catch (TException e) {
+                LOGGER.error("error processing request", e);
+                return;
+            } catch (RuntimeException e) {
                 try {
-                    FProtocol inputProto = inputProtoFactory.getProtocol(input);
-                    inputProto.setEphemeralProperties(request.ephemeralProperties);
-                    FProtocol outputProto = outputProtoFactory.getProtocol(output);
-                    processor.process(inputProto, outputProto);
-                } catch (TException e) {
-                    LOGGER.error("error processing request", e);
-                    return;
-                } catch (RuntimeException e) {
-                    try {
-                        conn.publish(request.reply, output.getWriteBytes());
-                        conn.flush(Duration.ofSeconds(60));
-                    } catch (Exception ignored) {
-                    }
-                    return;
+                    conn.publish(request.reply, output.getWriteBytes());
+                    conn.flush(Duration.ofSeconds(60));
+                } catch (Exception ignored) {
                 }
-
-                if (!output.hasWriteData()) {
-                    return;
-                }
-
-                // Send response.
-                conn.publish(request.reply, output.getWriteBytes());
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights
<details>
</details>

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
```diff
@@ -364,76 +364,97 @@ protected MessageHandler newRequestHandler() {
             Map<Object, Object> ephemeralProperties = new HashMap<>();
```
brettkail-wk, line 364: (Arbitrary line.)

FYI, this is a public repo, but the contents of that RFD link are not visible to all.

I'm wary of architecture where a request is sent to a NATS group (global), then routed to specific instance (local), and then routes back to a separate shared queue (global) if that instance is overloaded.  That broadly feels wrong.  A few questions that come to mind: What if the instance that received the message is too overloaded to put it back on the global queue?  What prevents the same problem from happening on the secondary global queue?  It seems like we're just pushing the problem around rather than addressing it (work is pushed to instances rather than pulled when they have capacity, requests aren't handled quickly enough, and there's not a good way to apply backpressure so that NATS chooses a different queue group member when a queue group member is overloaded).

I'm wary of this implementation that introduces implicit API that `FNatsServer` will always `.execute` a specific implementation class.

brendanbartels-wf: Yep, because this is a public repo, I didn't think the contents of that RFD are necessarily relevant, more just optional background to help exemplify potential. (There are [other PRs](https://github.com/Workiva/frugal/pull/1228) that reference RFDs, so my mistake, I thought there was precedent.)

The point of this PR was not to implement or justify that RFD, but just to expose a POJO to give consumers more flexibility, if that change has risks, and if those risks warrant attention. I'm happy to just discuss only the POJO to keep things in scope for this repo.

You mentioned you were wary of exposing this POJO class, which is fair, it is another published contract for consumers. I'm trying to sort through when it is okay to make something public and when it is not. What criteria do you think is good for that question?

brettkail-wk: Adding API makes it harder to change the impl, so I try to avoid unless there's a good use case :-).  When adding API, I try to avoid limiting future implementation.  Promising to pass a specific implementation of `Runnable` to `execute` is awkward, and exposing an inner (non-`static`) class is particularly odd.  In some ways, `FNatsServer.newRequestHandler` seems closer to the abstraction you want than `Runnable`...

brendanbartels-wf: Thank you for your feedback.

Another way to think of this PR is that it aims to improve the consumer's ability to interact with the `BlockingQueue` on the `ExecutorService`. The consumer can already provide their own instances of these classes and thus has full access to modify them, but is yet unable to do potentially useful operations (like examining or rearranging queued requests) since there is no interface over the data. So this PR is trying to fill that interface gap.

I will try to rework this branch to incorporate your suggestions.

brendanbartels-wf:
Closing for re-work.

lib/java/src/main/java/com/workiva/frugal/server/FNatsServer.java
brettkail-wk: Yeah, I understand that, I'm just wondering if `BlockingQueue`/`Executor` are the right level of abstraction.  You ideally want an integration point for "handle this Frugal-based NATS message" (sync/async, same/diff instance, etc.), but `Executor` only provides an integration point for "run arbitrary code", so the PR extends it to "run arbitrary code that always implements some other API to extract details about the code".  That's not a "clean" abstraction, but perhaps it's good enough if the use case justifies it.  I'll defer to others.

PR #1321, opened 2020-08-04
Dart: Add use_null_for_unset option

brettkail-wk:
### Story:
Commits are intended to be reviewed independently.  Notably, the 4th commit adds "baseline" files to `test/expected/dart.nullset` so that the changes added by the new option can actually be reviewed.

From the final commit message:
```
There are several issues with the current generation:
    1.  getter/setter methods are generated, which bloat the generated code
    2.  __isset_x fields are generated, which bloat the generated code and
        memory requirements.
    3.  isSet returns true for a field that has been explicitly set to null,
        which prevents validate() from detecting that a required field has
        not been set prior to writing.
```

### Acceptance Criteria:

### Design Notes:
Previous attempts (#1040, #1185) have attempted to solve the first two issues in a backwards compatible way, but there are subtle backwards incompatibilities.  Additionally, using `??` in the getter methods causes inconsistencies for non-primitive/reference fields (i.e., struct, map, list fields).  Instead, this PR adds a new option that can be enabled on a per-endpoint basis so that code can be incrementally tested with the new semantics.  See the commit message of the last commit for details.

The first three commits are improvements to the generated code that could be submitted independently if the 4th/5th commits are controversial.  Aside from those first three commits, no changes are intended to the generated output unless the new option is provided.

### How To Test:
Generate an API using `-gen dart:use_null_for_unset` and include in an application.

### My Test Results:
Despite the incompatibilities noted in that final commit, the changes have been tested on a large API+application, and it required no changes.

#### Reviewers:
@Workiva/product2 @brianshannan-wf 
@patkujawa-wf @robbecker-wf (involved in original 1040)
@fredkneeland-wf (author of 1185)
@kevinfox-wk @danielszopa-wk (reviewed the output of an earlier version)


aviary3-wk:
## Security Insights
<details>
</details>

## Action Items
---

test/expected/dart.nullunset/actual_base/f_thing.dart
```diff
@@ -133,10 +118,12 @@ class thing implements thrift.TBase {
...
-    oprot.writeFieldBegin(_AN_ID_FIELD_DESC);
-    oprot.writeI32(this.an_id);
-    oprot.writeFieldEnd();
-    if (this.a_string != null) {
+    if (isSetAn_id()) {
+      oprot.writeFieldBegin(_AN_ID_FIELD_DESC);
+      oprot.writeI32(this.an_id);
+      oprot.writeFieldEnd();
+    }
+    if (isSetA_string()) {
```
kevinfrommelt-wf: Prior to your change `an_id` was written without a null (or isSet) check, but now it is behind a null (or isSet) check. Is this a bug in the existing generator? Kind of seems like the existing generator has ways to unset the id field without nulling it out and it will still be written.

brettkail-wk: You're right, the current generator `write` method unconditionally writes default-required fields, even if `unset` is called.  The semantics of default-required are [vaguely specified](https://thrift.apache.org/docs/idl#default-requiredness-implicit), so I'm not sure the current generator is "wrong", but I do find the semantics somewhat confusing.  This section of the doc is particularly difficult to read:
> Although in theory these fields are supposed to be written ("req-out"), in reality unset fields are not always written. This is especially the case, when the field contains a value, which by definition cannot be transported through thrift. The only way to achieve this is by not writing that field at all, and that's what most languages do.

It seems like they were trying to say "fields default to unset, languages supporting null will probably use null for unset, and since there's no way to write out a null, those languages will skip writing unset fields", but they didn't want to set "null" since they were trying to be language agnostic?  Well, even if that's _not_ what they were trying to say, that's my interpretation, so that's what the new Dart generator implements :-).

brianshannan-wf: I agree with that interpretation. In general I think thrift poorly defined and implemented the default presence, and as we copied their generator output to be compatible, that means we poorly implemented it.

compiler/generator/dartlang/generator.go
```diff
@@ -802,20 +821,32 @@ func (g *Generator) generateFieldMethods(s *parser.Struct) string {
...
-		contents += fmt.Sprintf(tabtab+"this._%s = %s;\n", fName, fName)
-		if dartPrimitive {
-			contents += fmt.Sprintf(tabtab+"this.__isset_%s = true;\n", fName)
+		if !g.useNullForUnset() {
+			contents += g.generateFieldComment(field, tab)
+			contents += fmt.Sprintf(tab+"%s get %s => this._%s;\n\n", dartType, fName, fName)
+			contents += g.generateFieldComment(field, tab)
+			contents += fmt.Sprintf(tab+"set %s(%s %s) {\n", fName, dartType, fName)
+			contents += fmt.Sprintf(tabtab+"this._%s = %s;\n", fName, fName)
+			if !g.useNullForUnset() && dartPrimitive {
```
brianshannan-wf, line 830: nit: I think the `!g.useNullForUnset()` here is redundant

test/expected/dart.nullunset/variety/f_awesome_exception.dart
```diff
@@ -191,18 +160,23 @@ class AwesomeException extends Error implements thrift.TBase {
...
       oprot.writeString(this.reason);
       oprot.writeFieldEnd();
     }
-    oprot.writeFieldBegin(_DEPR_FIELD_DESC);
     // ignore: deprecated_member_use
-    oprot.writeBool(this.depr);
-    oprot.writeFieldEnd();
+    if (isSetDepr()) {
+      oprot.writeFieldBegin(_DEPR_FIELD_DESC);
+    // ignore: deprecated_member_use
```
brianshannan-wf, line 176: nit: It looks like the write method [ignores](https://github.com/Workiva/frugal/pull/1321/commits/ce94d2160831f24d93baa699ec117e3cfdc61725#diff-66f97854b47c41695e2116121a2ee428R1109) the passed in indentation, so this comment is at the wrong level. This obviously doesn't affect functionality and wasn't changed as part of this, so I don't think it absolutely needs fixed.

brianshannan-wf:
I like this approach much more than past approaches as we don't have to worry about forcing semantic changes that have previously caused issues in a non-major upgrade. Overall looks good to me.

test/expected/dart.nullunset/variety/f_awesome_exception.dart
brettkail-wk: Fixed, thanks.  I had noticed that, concluded it was pre-existing, and then forgot to circle back.

compiler/generator/dartlang/generator.go
brettkail-wk: Fixed, thanks.  I reworked this area of code many times as I was exploring the various edge cases around required/optional/default-required, primitive/non-primitive, default/no-default, and I must have overlooked this.

natewoods-wf:

fredkneeland-wf:
QA +1

- [x] At least one InfRe Squad 2 member has reviewed and +1'd
  - [x] DPC frugal code regenerated and tested with this branch: https://github.com/Workiva/doc_plat_client/pull/9671
  - [x] Passing skynet functional tests: https://wf-skynet-hrd.appspot.com/apps/test/smithy/2408681/1

PR #1337, opened 2020-08-31
SAPI-2496: Prepare for Thrift 0.13.0

natewoods-wf:
# [SAPI-2496](https://jira.atl.workiva.net/browse/SAPI-2496)
Issue Status

### Story:
~~Pulling in thrift 0.13.0 to patch CVE's.~~

~~CVE-2019-0210~~
~~CVE-2019-0205~~
~~CVE-2018-1320~~

Upstream thrift renamed `TApplicationException.read(iprot)` to `TApplicationException.readFrom(iprot)`.  So this effort introduces a SHIM class to encapsulate the logic so changes like upstream changes don't break downstream changes.

This should prepare us for a minimal update pr to update to thrift 0.13.0, which will likely just be a revert of the last commit on this PR with some follow up cleanup. 

Both this and the following PR will be minor tags (features added in semver).

### Acceptance Criteria:

### Design Notes:
Original Code, left here for comparison purposes

https://github.com/Workiva/frugal/blob/320fa6255d0b3a295bf297effa93853d1b000aa7/examples/java/src/main/java/v1/music/FStore.java#L106-L194

Current implementation, abstracted protocolFactory and transport out into a base class.  This is a new design is semi-consistent with upstream, but includes all the protocol and limiting things we have added to frugal in it's implementation.  Also, since our transport physically returns an object, we cannot split send and recv like thrift has in their generated logic, so this is probably as good as we can get for now. 

https://github.com/Workiva/frugal/blob/fc23cb925266fcdfc9ebe05851d21273eb3098d0/examples/java/src/main/java/v1/music/FStore.java#L106-L135

Upstream (thrift) chooses a common base class for generate clients to extend and use factories instead.  https://www.javadoc.io/doc/org.apache.thrift/libthrift/latest/org/apache/thrift/TServiceClient.html

```java
  public static class Client extends org.apache.thrift.TServiceClient implements Iface {
    public static class Factory implements org.apache.thrift.TServiceClientFactory<Client> {
      public Factory() {}
      public Client getClient(org.apache.thrift.protocol.TProtocol prot) {
        return new Client(prot);
      }
      public Client getClient(org.apache.thrift.protocol.TProtocol iprot, org.apache.thrift.protocol.TProtocol oprot) {
        return new Client(iprot, oprot);
      }
    }

    public Client(org.apache.thrift.protocol.TProtocol prot)
    {
      super(prot, prot);
    }

    public Client(org.apache.thrift.protocol.TProtocol iprot, org.apache.thrift.protocol.TProtocol oprot) {
      super(iprot, oprot);
    }

    public Album buyAlbum(java.lang.String ASIN, java.lang.String acct) throws PurchasingError, org.apache.thrift.TException
    {
      send_buyAlbum(ASIN, acct);
      return recv_buyAlbum();
    }

    public void send_buyAlbum(java.lang.String ASIN, java.lang.String acct) throws org.apache.thrift.TException
    {
      buyAlbum_args args = new buyAlbum_args();
      args.setASIN(ASIN);
      args.setAcct(acct);
      sendBase("buyAlbum", args);
    }

    public Album recv_buyAlbum() throws PurchasingError, org.apache.thrift.TException
    {
      buyAlbum_result result = new buyAlbum_result();
      receiveBase(result, "buyAlbum");
      if (result.isSetSuccess()) {
        return result.success;
      }
      if (result.error != null) {
        throw result.error;
      }
      throw new org.apache.thrift.TApplicationException(org.apache.thrift.TApplicationException.MISSING_RESULT, "buyAlbum failed: unknown result");
    }

    public boolean enterAlbumGiveaway(java.lang.String email, java.lang.String name) throws org.apache.thrift.TException
    {
      send_enterAlbumGiveaway(email, name);
      return recv_enterAlbumGiveaway();
    }

    public void send_enterAlbumGiveaway(java.lang.String email, java.lang.String name) throws org.apache.thrift.TException
    {
      enterAlbumGiveaway_args args = new enterAlbumGiveaway_args();
      args.setEmail(email);
      args.setName(name);
      sendBase("enterAlbumGiveaway", args);
    }

    public boolean recv_enterAlbumGiveaway() throws org.apache.thrift.TException
    {
      enterAlbumGiveaway_result result = new enterAlbumGiveaway_result();
      receiveBase(result, "enterAlbumGiveaway");
      if (result.isSetSuccess()) {
        return result.success;
      }
      throw new org.apache.thrift.TApplicationException(org.apache.thrift.TApplicationException.MISSING_RESULT, "enterAlbumGiveaway failed: unknown result");
    }

  }
```

Same idea, just different implementations.  I can transition my `helper` logic to be subclass based if desired.

### How To Test:
Automated tests should cover all verification of generated code changes and library changes.

### My Test Results:
Check CI status on last commit.

#### Reviewers:
@Workiva/service-platform

### DIFF
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>3.10.0-SNAPSHOT</version>
+    <version>3.9.9</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
@@ -50,7 +50,7 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.13.0</version>
+            <version>0.9.3</version>
         </dependency>
         <dependency>
             <groupId>io.nats</groupId>
@@ -181,7 +181,7 @@
                 <configuration>
                     <configLocation>checkstyle.xml</configLocation>
                     <consoleOutput>true</consoleOutput>
-
+                    
                     <failOnViolation>false</failOnViolation>
                     <includeTestSourceDirectory>true</includeTestSourceDirectory>
                     <includeResources>false</includeResources>
@@ -195,7 +195,7 @@
                             <goal>check</goal>
                         </goals>
                         <configuration>
-
+                            
                             <failOnViolation>false</failOnViolation>
                         </configuration>
                     </execution>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>3.9.9-SNAPSHOT</version>
+    <version>3.10.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>3.9.9</version>
+    <version>3.9.9-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
@@ -50,7 +50,7 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.9.3</version>
+            <version>0.13.0</version>
         </dependency>
         <dependency>
             <groupId>io.nats</groupId>
@@ -181,7 +181,7 @@
                 <configuration>
                     <configLocation>checkstyle.xml</configLocation>
                     <consoleOutput>true</consoleOutput>
-                    
+
                     <failOnViolation>false</failOnViolation>
                     <includeTestSourceDirectory>true</includeTestSourceDirectory>
                     <includeResources>false</includeResources>
@@ -195,7 +195,7 @@
                             <goal>check</goal>
                         </goals>
                         <configuration>
-                            
+
                             <failOnViolation>false</failOnViolation>
                         </configuration>
                     </execution>
--- a/lib/java/src/main/java/com/workiva/frugal/provider/FServiceClient.java
+++ b/lib/java/src/main/java/com/workiva/frugal/provider/FServiceClient.java
@@ -44,13 +44,8 @@ public class FServiceClient {
 	}
 
 	protected void requestBase(FContext ctx, String method, TSerializable args, TSerializable res) throws TException {
-		TMemoryOutputBuffer memoryBuffer = new TMemoryOutputBuffer(transport_.getRequestSizeLimit());
-		FProtocol oprot = protocolFactory_.getProtocol(memoryBuffer);
-		oprot.writeRequestHeader(ctx);
-		oprot.writeMessageBegin(new TMessage(method, TMessageType.CALL, 0));
-		args.write(oprot);
-		oprot.writeMessageEnd();
-		TTransport response = transport_.request(ctx, memoryBuffer.getWriteBytes());
+		byte[] payload = prepareMessage(ctx, method, args, TMessageType.CALL);
+		TTransport response = transport_.request(ctx, payload);
 		FProtocol iprot = protocolFactory_.getProtocol(response);
 		iprot.readResponseHeader(ctx);
 		TMessage message = iprot.readMessageBegin();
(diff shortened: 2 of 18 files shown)


aviary2-wf:
## Security Insights
<details>
</details>

## Action Items
---

lib/java/pom.xml
```diff
@@ -50,7 +50,7 @@
         <dependency>
             <groupId>org.apache.thrift</groupId>
             <artifactId>libthrift</artifactId>
-            <version>0.9.3</version>
+            <version>0.13.0</version>
```
brettkail-wk, line 53: The Frugal API base class / shim only has value if we introduce it separately from the libthrift upgrade.  As is, this PR would mean services can't incrementally upgrade API dependencies (the service must upgrade to new Frugal to pull in the base class before any API dependency can be upgraded, but that can't be done due to the libthrift API incompat unless all API dependencies are upgraded).  Suggest:
1. add new base class but using libthrift 0.9.3 and release frugal version B (current frugal version A does not contain the shim)
2. services and API dependencies can upgrade/release incrementally (codegen from A runs with runtime A or B, codegen from B only runs with runtime B, but services can upgrade without issue)
3. after some period we can upgrade to libthrift 0.13.0 and release version C (codegen from A now broken, codegen from B or C runs with runtime from B or C).

lib/java/src/main/java/com/workiva/frugal/provider/FServiceClient.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.thrift.transport.TTransportException;
+import org.apache.thrift.TSerializable;
+
+/**
+ * FProtocolHelper is used by generated code to consolidate the logic generated.
+ */
+public class FServiceClient {
+
+	protected FTransport transport_;
+	protected FProtocolFactory protocolFactory_;
```
brettkail-wk, line 39: Does generated code still reference these?  If not, suggest `private`.

The trailing `_` is not the prevailing style in this repo.

natewoods-wf:
@aviary-wf want to re-scan, hopefully I have all the POM's up to thrift 0.13.0 now.

lib/java/src/main/java/com/workiva/frugal/provider/FServiceClient.java
natewoods-wf: all protected and `_` naming conventions come from upstream: https://github.com/apache/thrift/blob/af7ecd6a2b15efe5c6b742cf4a9ccb31bcc1f362/lib/java/src/org/apache/thrift/TServiceClient.java#L40

Willing to change if desired.

brettkail-wk: It seems strange to keep them protected, but perhaps it makes codegen changes in the future easier since they can directly access the variables.  I'm fine with it.

It looks like their convention is to use trailing underscores for code that interops with generated code, and the Frugal codegen just didn't adopt that convention (presumably naming a method param `protocol` or `protocolFactory` would break things, but just no one does that, and it's presumably thought to be unlikely that someone names a param with trailing underscores).  I'm fine with it.

lib/java/pom.xml
natewoods-wf: Hey @brettkail-wk , we have met as a messaging team and have decided to move forward with the following rollout process: We will cut this as a stand alone minor version of frugal and will have a similar consuming pr of `messaging-sdk` (internal).  At this point, teams can consume the bump to `thrift` 0.13.0 in one PR, by updating the frugal generated code using the latest tool, the java frugal library to match the generator version, prune any previous `libthrift` exclusions from the `pom.xml` files, and updating `frugal`, `thrift` and `messaging-sdk` versions accordingly.

Given that is this fallout of a security request, we feel this will be the least time consuming and least disruptive process to product teams with only one PR per repo to consume the required new version of thrift.  Additionally, this should remove the necessity of our team maintaining 4 separate `thrift*` forks.  If you have any concerns with this rollout strategy, feel free to reach out to our DM and PM (@shannonharder-wk and @erikpetersen-wf).

brianshannan-wf: I agree with Brett. The versioning guidance in the past has been "library version >= generator version works" (within the same major version). Breaking from this requires dependencies (and dependencies of dependencies, etc.) to be updated in sync, which, from experience, can be a large pain.

What's the point of introducing the shim if everyone is going to be upgraded in one go? If that is going to happen, you could leave the generated code alone except for changing the method called to the one in thrift 0.13.0 and make no other changes.

natewoods-wf: we cannot leave the generated code alone as thrift 0.13.0 has renamed `TApplicationException.read` to `TApplicationException.readFrom` which means no matter what, there needs to be a generator + library change at the same time.

brianshannan-wf: I agree, which I thought was the point of the shim, that it would make this a more seamless rollout. But if this change is going to happen in one release, it seems the shim isn't doing anything and the change to the generated code just has to be `read` -> `readFrom`? Or maybe I'm missing something?

natewoods-wf: The shim is just there to help if there are similar problems in future versions of thrift and to make frugal generated code more closely resemble thrift generated code.  And, given the regeneration is taking place anyway (and again, this is a security compliance concern) then we figured it would be a good time to roll out the shim as well.

davecasleton-wf: What is the timeline for the security issue needing to be resolved?

If this is going to require a mass coordinated deployment across very many services, it seems like the least palatable approach. That will inevitably block work across very many teams in order to save a few forks managed by one.

natewoods-wf: After further discussion [in-org-slack] we have decided to split this into a shim minor version that references thrift 0.9.3's `read` method and something to patch the lack of `TSerializable` (added in thrift 0.10.0), and another minor version that updates the shim and library to thrift 0.13.0 and `readFrom`.  Suggestions for the public is to move to the version using thrift 0.13.0 asap given the mentioned CVE issues, but splitting this release into to minor version enables our product teams a longer rollout path to adopt this library change if they choose to do so.

natewoods-wf:
Apologies for the force pushing of `rollback(thrift: 0.13.0 -> 0.9.3 :arrow_down:` as I'm trying to simplify the follow up PR creation.

natewoods-wf:

PR #1050, opened 2018-11-19
INFENG-6690: GopherJS Compile Target

natewoods-wf:
# [INFENG-6690](https://jira.atl.workiva.net/browse/INFENG-6690): GopherJS Compile Target

- [ ] At least one InfRe Squad 2 member has reviewed and +1'd
- [ ] Code has been tested and results documented
- [ ] Updates to documentation and change logs if necessary
- [ ] Verify and document changes to any other Messaging components

 ### Story
TL;DR: transpiled frugal code (that imports a constant) creates an 8MB JS application; let’s not do that.

GopherJS is a tool that allows the compilation of go code to javascript in order to share critical application logic between clients and servers.  This compilation from go to js is referred to as `transpiling` for the remainder of this document.  This process of transpiling application code has become a critical component of go application development enabling complex logic to be written once, but used by both clients and servers. As such, we would like to provide better transpilation performance with regards to generated code size.

While working with gopherjs for the past few years, we have identified imports as the primary reason for large transpiled applications.  Due to the way imports are used in go, gopherjs is unable to “shake” import trees, and thus must generate all public types and types referenced in package `init`s within the entire import tree.  Effects of import optimization can be found in a similar series of changes made here: .https://github.com/gopherjs/websocket/pull/20.

Currently, thrift and frugal import net/http which brings in an entire http serving library along with TLS encryption and many other fairly large packages that are unnecessary when running an application in javascript.  To mitigate these import trees, thrift and frugal go libraries have been vendored and modified to import less things.  Additionally, where possible, the build flag `// +build !js` has been added to things not needed at all in client logic, such as `simple_server.go` along with a few others.  Finally, a few isomorphic functions have been put in place in order to enable smaller generation of transpiled code.

### Acceptance Criteria
TODO: This space intentionally left blank for @Workiva/product2 members to fill out.

### Design Notes
To accommodate for gopherjs being a first class consumer of frugal, we should make a new compile target for it that allows us to optimize the resulting code.  Currently, this target configures some options and essentially runs the existing golang compiler.  This should get us started quickly and allow us to extend this new target in the future.

As noted above, imports can drastically increase the size of a transpiled application.  As such, a new frugal library and vendored copy of thrift packages have been included in this directory that allows us to squeeze further transpiling performance out of them.

[...]

PR #1352, opened 2020-09-11
SPLAT-65: GO: encapsulate Flush (thrift 0.13.0 preparation)

natewoods-wf:
# [SPLAT-65](https://jira.atl.workiva.net/browse/SPLAT-65)
Issue Status

This should make moving to thrift 0.13.0 a little easier

### Story:
Similar to https://github.com/Workiva/frugal/pull/1337, but slightly more complicated with the target branches, given that we have tagged v3.10.0 and v3.11.0 already.

Thrfit 0.13.0 converts [TProtocol's `Flush()` to `Flush(ctx)`](https://github.com/apache/thrift/blob/v0.13.0/lib/go/thrift/protocol.go#L78) (where ctx is a `context.Context`)(changes can be seen here https://github.com/Workiva/frugal/pull/1350).  This effort aims to internalize `Flush` calls into the frugal library, such that generated code can be backwards and forwards compatible with thrift 0.13.0 and 0.9.3.  This change is targeting 0.9.3, and will allow the update to thrift 0.13.0 be a library only change, not affecting the generated code.

```
# Another chagne in thrift 0.13.0
have GetTransport(thrift.TTransport) thrift.TTransport
want GetTransport(thrift.TTransport) (thrift.TTransport, error)
# We should be able to just make this change as the only function with this signature is not used outside this repo.
# Searched our orgs sourcegraph: `NewTFramedTransport -file:(^|/)vendor/ lang:go`
# Or searching all of github: https://github.com/search?p=2&q=frugal.NewTFramedTransport&type=Code
```

To encapsulate some of the work a `frugal.FClient` has been added (similar to Java's FClient added in #1337) that encapsulates some of the lower level send/receive/publish methods.  This is generally not for external use and should only be consumed by generated code.

While making these changes, minor cleanup to publish logic was included to make generated code somewhat more readable.  While this was not necessary explicitly, it allowed for `TStruct` to be the primary interface between generated clients and the actions performed with them.

### Acceptance Criteria:
- [ ] At least one Service Platform member has reviewed and +1'd

### Design Notes:
FClient is similar to thrift's TClient, but includes all the other fun things that frugal can do: https://github.com/apache/thrift/blob/v0.13.0/lib/go/thrift/client.go#L8

Follow up questions:

1. Do we need `Open()` or `Close()` on Publisher transports?  The Nats and Stomp implementations don't really do much, and are better managed by the parent provider, and not in generated logic IMO.
2. Can we update subscribers to use the same helpers the publishers now use?
3. Can we move to a single dependency management solution for go, this project now has, godep, glide and module support?

### How To Test:
```sh
cd examples/go
make install
rm -rf vendor/git.apache.org # updated to 0.13.0 on my machine and couldn't figure out how to rollback
go get git.apache.org/thrift.git
pushd $GOPATH/src/git.apache.org/thrift.git
git checkout 0.9.3
popd
go run ./httpServer &
go run ./httpClient
```

### My Test Results:

```sh
# Server
$ go run ./httpServer/                                     
Starting the http server...
WARN[0041] Deprecated function 'Store.EnterAlbumGiveaway' was called by a client

# Client
 $ go run ./httpClient/
==== CALLING *music.FStoreClient.buyAlbum ====
==== CALLED  *music.FStoreClient.buyAlbum ====
Bought an album Album({Tracks:[Track({Title:Comme des enfants Artist:Coeur de pirate Publisher:Grosse Boîte Composer:Béatrice Martin Duration:169 Pro:ASCAP})] Duration:1200 ASIN:c54d385a-5024-4f3f-86ef-6314546a7e7f})
WARN[0000] Call to deprecated function 'Store.EnterAlbumGiveaway'
==== CALLING *music.FStoreClient.enterAlbumGiveaway ====
==== CALLED  *music.FStoreClient.enterAlbumGiveaway ====
```

#### Reviewers:
@Workiva/service-platform

### DIFF
--- a/compiler/generator/golang/generator.go
+++ b/compiler/generator/golang/generator.go
@@ -1326,8 +1326,8 @@ func (g *Generator) GeneratePublisher(file *os.File, scope *parser.Scope) error 
 	}
 
 	publisher += fmt.Sprintf("type %sPublisher interface {\n", scopeCamel)
-	// publisher += "\tOpen() error\n"
-	// publisher += "\tClose() error\n"
+	publisher += "\tOpen() error\n"
+	publisher += "\tClose() error\n"
 	for _, op := range scope.Operations {
 		publisher += fmt.Sprintf("\tPublish%s(ctx frugal.FContext, %sreq %s) error\n", op.Name, args, g.getGoTypeFromThriftType(op.Type))
 	}
@@ -1351,6 +1351,9 @@ func (g *Generator) GeneratePublisher(file *os.File, scope *parser.Scope) error 
 	}
 	publisher += "\treturn publisher\n"
 	publisher += "}\n\n"
+
+	publisher += fmt.Sprintf("func (p %sPublisher) Open() error { return p.client.Open() }\n", scopeLower)
+	publisher += fmt.Sprintf("func (p %sPublisher) Close() error { return p.client.Close() }\n\n", scopeLower)
 
 	prefix := ""
 	for _, op := range scope.Operations {
--- a/compiler/generator/golang/generator.go
+++ b/compiler/generator/golang/generator.go
@@ -1687,7 +1687,7 @@ func (g *Generator) generateClient(service *parser.Service) string {
 	} else {
 		contents += "\tc       frugal.FClient\n"
 	}
-	contents += "\tmethods         map[string]*frugal.Method\n"
+	contents += "\tmethods map[string]*frugal.Method\n"
 	contents += "}\n\n"
 
 	contents += fmt.Sprintf(
@@ -1699,7 +1699,7 @@ func (g *Generator) generateClient(service *parser.Service) string {
 		contents += fmt.Sprintf("\t\tF%sClient: %sNewF%sClient(provider, middleware...),\n",
 			service.ExtendsService(), g.getServiceExtendsNamespace(service), service.ExtendsService())
 	} else {
-		contents += "c:    frugal.NewFStandardClient(provider),\n"
+		contents += "\tc:    frugal.NewFStandardClient(provider),\n"
 	}
 	contents += "\t\tmethods:         methods,\n"
 	contents += "\t}\n"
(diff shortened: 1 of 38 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

compiler/generator/golang/generator.go
```diff
@@ -1303,15 +1301,6 @@ func (g *Generator) generateImportProtection(include *parser.Include) string {
 
 // GenerateConstants generates any static constants.
 func (g *Generator) GenerateConstants(file *os.File, name string) error {
```
ryanwilson-wk, line 1303: Is this function still needed if it always returns `nil` now?

ryanwilson-wk (approved):
`+1,580 -8,600`... looks great to me. I can't say I fully understand all the details, but it looks similar to what we started on the Java end.

charliekump-wf (approved):
It looks good to me as well, expecially the amount of generated code removed. I don't have any strong feelings on your follow up questions except it looks to me the Open()/Close() are unnecessary and a single go dep manager 👍 👍

compiler/generator/golang/generator.go
natewoods-wf: Yes, any language Generator must have all the functions on the `LanguageGenerator` interface.

https://github.com/Workiva/frugal/blob/ce94d2160831f24d93baa699ec117e3cfdc61725/compiler/generator/generator.go#L131

natewoods-wf:
Updating gopherjs included all the previous changes to the go library since https://github.com/Workiva/frugal/pull/1050 that have not been moved to gopherjs as well.  If this seems like too much to do in this PR, let me know and I can split that update to a separate PR.  But generally speaking, lib/gopherjs should be a near copy of lib/go but with correlation ID generation being cleaner to transpile.

charliekump-wf (approved):
+1

natewoods-wf:
Further testing can be performed on the integration branch.

@Workiva/release-management-p for merge into v1.10.0.1-dev

rmconsole-wf (approved):
+1 from RM

PR #1360, opened 2020-09-18
3.10.X Changes for Java

ryanwilson-wk:
### Story:
https://jira.atl.workiva.net/browse/SPLAT-108

### Acceptance Criteria:

### How To Test:
Added a super service that the FrugalTest service depends on and verified it wasn't working prior to our changes

`wk test run cross-local` BOOM! (might just work for me though :()

### My Test Results:
All tests passed

#### Reviewers:
@Workiva/service-platform


aviary3-wk:
## Security Insights

## Action Items
---

natewoods-wf (approved):
FYI @brettharper-wf

compiler/generator/java/generator.go
```diff
@@ -2686,7 +2686,7 @@ func (g *Generator) generateClient(service *parser.Service, indent string) strin
 		contents += indent + fmt.Sprintf("public static class Client extends %s.Client implements Iface {\n\n",
 			g.getServiceExtendsName(service))
 	} else {
-		contents += indent + "public static class Client implements Iface {\n\n"
+		contents += indent + "public static class Client extends FServiceClient implements Iface {\n\n"
```
brettkail-wk, line 2689: This seems like a mistake to me since the whole point of the separate `InternalClient` API is to hide implementation details.  The only reason this is needed is for `service A extends B`, where the `InternalClient` of A needs to somehow access the new helpers.

Unfortunately, the helpers were defined as member variables in `FServiceClient` rather than in static methods.  Two ideas come to mind.

### 1. Create a synthetic FServiceClient
If `service A extends B`, then generate a special `InternalClient`:
```
public static class InternalClient extends B.Client {
  private final FServiceClient serviceClient;

  public InternalClient(FServiceProvider provider) {
    super(provider);
    this.serviceClient = new FServiceClient(provider);
  }

  public void m(FContext ctx, ...) {
    ...
    serviceClient.requestBase(...);
  }
}
```

Unfortunately, the helper methods are `protected`, so we have to create a synthetic class just for visibility:
```
public static class InternalClient extends B.Client {
  private static class InternalServiceClient extends FServiceClient {
    InternalServiceClient(FServiceProvider provider) {
      super(provider);
    }

    protected requestBase(...) { return super.requestBase(...); }
  }

  private final InternalServiceClient serviceClient;

  public InternalClient(FServiceProvider provider) {
    super(provider);
    this.serviceClient = new InternalServiceClient(provider);
  }

  ...
}
```

This extra inner class is rather unfortunate.

### 2. Introduce an FServiceClientUtils
The previous is complex.  Add a new `FServiceClientUtils` with static methods, and move all the guts of `FServiceClient` into that, and have it call those methods.  Then, in the `service A extends B` case, generate a special `InternalClient`:
```
public static class InternalClient extends B.Client {
  private FTransport transport_;
  private FProtocolFactory protocolFactory_;

  public InternalClient(FServiceProvider provider) {
    super(provider);
    this.transport_ = provider.getTransport();
    this.protocolFactory_ = provider.getProtocolFactory();
  }

  public void m(FContext ctx, ...) {
    ...
    FServiceUtils.requestBase(transport_, protocolFactory_, ...);
  }
}
```

This adds another ABI that is just used for internal clients, which means that the new `FServiceClientUtils` also needs to be added+released as 3.12.  However, it has some advantages:
1. it avoids bleeding the `FServiceClient` impl details into `Client`
2. it avoids generating yet another inner class from approach 1
3. it's otherwise backwards compatible with 3.10.0 for non-`extends` (because `FServiceClient` is still used for those use-cases, and that just forwards to `FServiceClientUtils`).

ryanwilson-wk: Yeah, I agree that this solution isn't ideal, and there is code smell by having both the `Client` and `InternalClient` both inherit from the `FServiceClient`. I think the main issue is that the `InternalClient` needs to implement the `Iface` in order to invoke the middleware wrapper, so it either needs to be able to extend the `InternalClient` from the super class or reimplement all the methods to satisfy the contract.

What do you think about renaming `InternalClient` to `RawClient` and having it be a public class? Then the subclass `RawClient` could inherit the superclass' `RawClient`, and the main `Client` that does the middleware wrapping doesn't need to know about the `FServiceClient`.

brettkail-wk: > I think the main issue is that the InternalClient needs to implement the Iface in order to invoke the middleware wrapper, so it either needs to be able to extend the InternalClient from the super class or reimplement all the methods to satisfy the contract.

...or `Sub.InternalClient` continues to extends `Super.InternalClient` as it does today as a way of getting an arbitrary implementation of the methods.  I don't think the actual implementation of the base methods of `InternalClient` actually matters since `Sub.Client extends Super.Client` without override base methods, so `Super.Client` will always invoke `Super.InternalClient` methods anyway.

> What do you think about renaming InternalClient to RawClient and having it be a public class

I think it's unfortunate that the `RawClient` detail has to be exposed, and it's unfortunate that it will clutter everyone's autocomplete.

I think I still lean towards `FServiceClientUtils` static methods.

ryanwilson-wk: Spoke with @brettkail-wk offline, and we agreed that the lesser of all evils would be to keep the name `InternalClient`, keep it public, but mark it with the `@Deprecated` flag to discourage external consumption.

compiler/generator/java/generator.go
```diff
@@ -2762,11 +2762,12 @@ func (g *Generator) generateAsyncClientMethod(service *parser.Service, method *p
 
 func (g *Generator) generateInternalClient(service *parser.Service, indent string) string {
 	contents := ""
+	contents += indent + "@Deprecated\n"
 	if service.Extends != "" {
-		contents += indent + fmt.Sprintf("private static class InternalClient extends %s.Client implements Iface {\n\n",
+		contents += indent + fmt.Sprintf("public static class InternalClient extends %s.InternalClient implements Iface {\n\n",
```
brettkail-wk, line 2767: It occurred to me over the weekend that this is going to generate code that is incompatible with _generated_ code for all current release (both 3.10.0 and 3.11.0), which is quite unfortunate.  That's particularly annoying since the primary use case here is for `BaseService`, and "vendoring" isn't widely used for Java, so every x-api repo generates+packages their own `BaseService` classes, which means that services that pull in many x-api repos pull in many copies of `BaseService`, which means you either have to play/rely on classpath ordering tricks or you have to freeze until everything can be upgraded (and the whole point of the 3.10.x release was so that we wouldn't have to freeze everything while we upgrade).

I know we agree the design choice of `InternalClient extends Client` isn't great (it's basically just abusing the type hierarchy to get arbitrary implementations of unused methods), and I know that the options [here](https://github.com/Workiva/frugal/pull/1360#discussion_r491112648) aren't great, but I'm wondering again if that isn't still better than this incompatibility-within-an-incompatibility...

ryanwilson-wk: Ok, I think I found a solution that checks all the boxes. I created a new interface called `Service` that represents the subclass and the `InternalClient` implements that instead of the full `Iface`. This allows us to:

- Keep backwards compatibility with 3.9
- Keep the `InternalClient` class `private`
- Better represent what the `InternalClient` is actually implementing
- Not mix the `InternalClient` and `Client` inheritance hierarchy unnecessarily

The only downside is the addition of a new interface, which I tried to make `private` but the middleware proxy generator didn't like that. What do you think about this solution @brettkail-wk?

brettkail-wk: Overall, this seems rather elegant.

> The only downside is the addition of a new interface, which I tried to make private but the middleware proxy generator didn't like that. 

This is indeed unfortunate, particularly that `Service` looks like something people should use but they probably shouldn't.  Two thoughts:
1. rename to `InternalIface`?  this one is harmless enough that I don't know that we have to do the same deprecated dance, but perhaps add javadoc that says something like "Contains only the methods defined directly by the service"?
1. this interface is only required for `InternalClient` of subservices, so only generate it in that case?  i.e., `InternalClient extends Iface` for base and `InternalClient extends InternalIface` for others?

ryanwilson-wk: 1 seems reasonable to me. For 2 though, that would dramatically increase the branching complexity of the generator. The `Client` has a member variable where the type would then have to change in a couple spots depending if it was a subclass or standalone, along with the call to the middleware wrapper and changing the type of the `InternalClient` as well. I think it would be cleaner to always generate the `InternalIface` so we don't introduce a bunch of conditionals. Then we just isolate it to one check when generating to `Iface` to see if we need to extend the superclass.

brettkail-wk: I'm willing to trade increased generator complexity for decreased generated code complexity, particularly since it would mean that code not using `extends` remains unchanged after this PR.  The javadoc for `InternalIface` is also quite awkward since there's no good way to justify that for services that don't extend anything.  Well, if you feel strongly, then go ahead.

ryanwilson-wk: To me the justification is consistency in architecture. You don't have to try to figure it which way you are doing based on if it is a subclass or not. I think given that and the added branching logic I would prefer to just always generate it. Probably just a matter of opinion.

brettkail-wk (approved):

natewoods-wf (approved):

natewoods-wf:
@Workiva/release-management-p for merge into integration branch

rmconsole-wf (approved):
+1 from RM

PR #1463, opened 2021-02-22
Drop gopherjs generation support

natewoods-wf:
### Story:
The GopherJS Compile target was added as an experimental target in https://github.com/Workiva/frugal/pull/1050.
It allowed our teams to generate client code without importing `net/http` which caused dramatic bloat when the browser application we were hosting was not serving any http handlers.
Recently, the final internal efforts to cleanup usage of these packages has been merged and as such, this logic can be safely removed.  This cleanup effort was also driven by the thrift CVEs over the past few years, whose fixed needed to be ported into our forked copy of thrift, causing un-needed support on our messaging platform teams.

### Acceptance Criteria:

### Design Notes:
It's a sad day to see this code of old be put out to pass.  But the good news is that is served us well for 3+ years and has done it's job well in reducing the overall compiled binary sizes.

### How To Test:
Verified no Workiva repositories generated with `-gen gopherjs` using sourcegraph and ensured no committed code imported the `github.com/Workiva/frugal/lib/gopherjs/*` packages.

```
# Source Graph Queries
github.com/Workiva/frugal/lib/gopherjs count:1000 -file:(^|/)vendor/ -repo:^github\.com/Workiva/frugal$ 
gen gopherjs count:1000 -file:(^|/)vendor/ -repo:^github\.com/Workiva/frugal$ 
```

### My Test Results:
The matching results are false positives or have been otherwise PR'd against.  Workiva teammates will see their PRs linked in the conversation thread.

#### Reviewers:
@Workiva/service-platform


aviary3-wk:
## Security Insights

## Action Items
---

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/Workiva/frugal
 go 1.15
 
 require (
+	github.com/Workiva/frugal/lib/go v0.0.0-20210120174923-290564696faf
+	github.com/apache/thrift/lib/go/thrift v0.0.0-20210120171102-e27e82c46ba4 // indirect
```
joshprzybyszewski-wf, line 7: 🔍  This seems :sus: - why was it added?

compiler/generator/generator.go
```diff
@@ -49,18 +49,14 @@ type LanguageOptions map[string]Options
...
+		"package_prefix":              "Package prefix for generated files",
+		"async":                       "Generate async client code using channels",
+		"use_vendor":                  "Use specified import references for vendored includes and do not generate code for them",
+		"slim":                        "Generate slim type definitions (WARNING: code generated by this may break code consumers, protocol logic should not change)",
 		"suppress_deprecated_logging": "Suppress decrecated API usage warning logging",
 	},
-	"gopherjs": Options{
-		"package_prefix": "Package prefix for generated files",
-		"use_vendor":     "Use specified import references for vendored includes and do not generate code for them",
-	},
```
joshprzybyszewski-wf, line 63: 🔍  This removes functionality with no replacement. Are we prepared to release v4 of the frugal binary?

lib/gopherjs/README.md
```diff
@@ -1,87 +0,0 @@
-# GopherJS Frugal Compile Target
-
-### Story
-TL;DR: transpiled frugal code (that imports a constant) creates an 8MB JS application; let’s not do that.
```
joshprzybyszewski-wf, line 4: oh if only it were 8MB :badpokerface:

Honestly, this PR is removal of publicly accessible packages. Are we prepared to cut frugal v4?

compiler/generator/generator.go
natewoods-wf: With the breaking API changes from downstream thrift in https://github.com/Workiva/frugal/pull/1458, I think it would make sense to cut a v4 with both of those PRs.

lib/gopherjs/README.md
natewoods-wf: With the breaking API changes from downstream thrift in https://github.com/Workiva/frugal/pull/1458, I think it would make sense to cut a v4 with both of those PRs.

go.mod
natewoods-wf: Good catch, my environment sporadically adds these files. I'll get those updated.

compiler/generator/generator.go
aaronstgeorge-wf: This would change the dart version as well. We can do that, but just to remind you dependency overrides are not allowed in dart production repos. A major requires bumping all repos that rely on frugal to rely to have a version that spans two majors, then running through again after the major flows through and tightening the range [link to official policy](https://github.com/Workiva/ApplicationFrameworks/blob/master/WdeskVersioning.md). Two passes on a good chunk of our dart repos is a non trivial amount of work for no breaking changes in dart.

joshprzybyszewski-wf: @aaronstgeorge-wf Agreed: There exists a lot of unnecessary pain in needing to support semver for multiple languages in the same repo. Major versions are scary for this reason, and perhaps we should be considering alternative solutions to hosting all languages in the same repo.

aaronstgeorge-wf: I actually forgot... Frugal is a public repo. Our semver contract previously has been that the major is the wire protocol, the minor is the library version, and the patch is just the patch. Given that this is public I think we should keep to that same standard.

natewoods-wf: Also, this code is vulnerable to not just one, but multiple thrift CVE's. So ideally IMO, it would be removed completely.

joshhayes-sheen-wf (approved):
Screen Shot 2021-02-24 at 2 52 47 PM

LGTM (Also, we don't use gopherjs)

natewoods-wf:
After searching both our internal sourcegraph for consumers, as well as pkg.go.dev ability to check for importers of a package, we have found that there are zero consumers of gopherjs's thrift or frugal packages currently.  Thus, I believe this to be a "safe" removal.  As for the version number this code change predicates, that discussion will be had on the release PR for merging `develop` into `master`.

https://pkg.go.dev/github.com/Workiva/frugal/lib/gopherjs/frugal?tab=importedby
https://pkg.go.dev/github.com/Workiva/frugal/lib/gopherjs/thrift?tab=importedby

compiler/generator/generator.go
joshprzybyszewski-wf: I'm not a regular frugal contributor, but as a consumer, it seems like there should be a fourth "semver" prepended to the standard three: `vWire.vMajor.vMinor.vPatch`. `vWire` would change maybe once every 10 years, `vMajor` could change a few times a year, and the lower Vs could change even more frequently.

This PR is clearly [a major change](https://semver.org/) (removal of publicly accessible code in a backwards incompatible way), but I won't pretend to have the "frugal cred" to force a frugal major version.

joshprzybyszewski-wf (approved):
I wish that `frugal` was in a place to release this as an actual major, but I think that getting rid of code we don't want to support (and we don't think is actually in use) is a good thing

nathancoleman-wf (approved):
Count me in favor of releasing a major version. Always great to remove unnecessary overhead. Thanks for doing this work @natewoods-wf

houstonking-wf (approved):

natewoods-wf:

rmconsole-wf (approved):
+1 from RM

PR #1608, opened 2021-06-23
java: Fix spurious IllegalArgumentException with NATS 2.2.0 and no responders

brettkail-wk:
### Story:
Commits are intended to be reviewed independently.  See commit messages for details.

I was developing an integration test issue where a Frugal request was being sent to a subject with no responders, and I was receiving the following exception in the logs:
```
java.lang.IllegalArgumentException: 4 > 0
	at java.base/java.util.Arrays.copyOfRange(Arrays.java:4029)
	at com.workiva.frugal.transport.FNatsTransport$Handler.onMessage(FNatsTransport.java:147)
	at io.nats.client.impl.NatsDispatcher.run(NatsDispatcher.java:98)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

After debugging, I discovered that I was running NATS 2.2.0, which includes nats-server PR 1478, which automatically sends a no-data status message with headers if there are no subscribers to a PUB.  nats.java PR 362 parses this and unconditionally delivers a no-data `Message`.

### Acceptance Criteria:

### Design Notes:
The first commit ignores status messages sent to the inbox since they are not valid Frugal messages.

The remaining commits adds support to immediately fail requests rather than waiting on a timeout if a 503 (NO_RESPONDERS) status message is sent to the Frugal inbox.  Since the message has no additional context, the inbox subscription is changed to a wildcard, and the op id is encoded in the reply subject.  In theory, the same approach could be used to remove the [transport response header decoding](https://github.com/Workiva/frugal/blob/eb8279149329b68bc5156b41e16e204ef6614afb/lib/java/src/main/java/com/workiva/frugal/transport/FAsyncTransport.java#L140-L141), which is somewhat expensive just to find the op id, but I chose to keep this PR simple for now.

I don't have a convenient setup for testing Go, Dart, or Python, so I have not checked whether status messages similarly cause issues for those languages, and I have not attempted to short-circuit the 

### How To Test:
Set up a client/server using NATS 2.1.9 and NATS 2.2.0, and send oneway/request to subjects with and without subscribers.

### My Test Results:
I tested the changes with an internal service that makes NATS requests and verified both the success and failure paths.

As part of the testing, I also had two servers sending requests to eachother, one with these updates and one without, so there should not be any interoperability problems.

#### Reviewers:
@Workiva/service-platform
FYI @scottmartin-wk

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
@@ -139,8 +139,12 @@ public class FNatsTransport extends FAsyncTransport {
     protected class Handler implements MessageHandler {
         public void onMessage(Message message) {
             try {
-                byte[] frame = message.getData();
-                handleResponse(Arrays.copyOfRange(frame, 4, frame.length));
+                if (message.isStatusMessage()) {
+                    LOGGER.debug("Received status message: {}", message);
+                } else {
+                    byte[] frame = message.getData();
+                    handleResponse(Arrays.copyOfRange(frame, 4, frame.length));
+                }
             } catch (TException e) {
                 LOGGER.warn("Could not handle frame", e);
             }
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
@@ -109,7 +109,7 @@ public class FNatsTransport extends FAsyncTransport {
             throw new TTransportException(TTransportExceptionType.ALREADY_OPEN, "NATS transport already open");
         }
         dispatcher = conn.createDispatcher(new Handler());
-        dispatcher.subscribe(inbox);
+        dispatcher.subscribe(inbox + ".*");
     }
 
     /**
@@ -126,6 +126,15 @@ public class FNatsTransport extends FAsyncTransport {
 
     @Override
     protected void flush(byte[] payload) throws TTransportException {
+        flush(null, payload);
+    }
+
+    @Override
+    protected void flushOp(long opId, byte[] payload) throws TTransportException {
+        flush(inbox + '.' + opId, payload);
+    }
+
+    private void flush(String inbox, byte[] payload) throws TTransportException {
         Status status = conn.getStatus();
         if (!isOpen(status)) {
             throw getClosedConditionException(status, "flush:");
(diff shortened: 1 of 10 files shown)


aviary-wf:
## Security Insights
<details>
</details>

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java
```diff
@@ -109,7 +109,7 @@ public void open() throws TTransportException {
             throw new TTransportException(TTransportExceptionType.ALREADY_OPEN, "NATS transport already open");
         }
         dispatcher = conn.createDispatcher(new Handler());
-        dispatcher.subscribe(inbox);
+        dispatcher.subscribe(inbox + ".*");
```
danielharasymiw-wf, line 112: I'm not terribly familiar with the inner workings of nats. Is this going to cause an issue if people are using this with a different service using a different version of this library? For example, older client sending a message to "inbox" but this newer server not listening on "inbox" but instead "inbox.*"

danielharasymiw-wf: @natewoods-wf

danielharasymiw-wf: Believe this was answered for me in the slack thread.

brettkail-wk: But to be sure, it does not.  The inbox string is a private implementation detail of the client.  All the servers in this repo (Java, Go, Python aio/torpedo) treat it as an opaque string.  So long as the client is subscribed to the string, it shouldn't matter which inbox string is used.

I've updated the "My Test Results" to mention that I did test the scenario of servers with/without these changes sending requests to eachother.

natewoods-wf (approved):
Hey @brettkail-wk!
We just ran into this issue the other day, so thanks for getting a patch up quick!  I appreciate the allocation of the error numbers in the non-java languages as that should make it more difficult to mess those up as future errors are added.  Our team has written tickets for updating the other languages and will be prioritizing them appropriately with our other work.

It took me a bit to reason about, but the way these inboxes are structured should be no concern when integrating against older versions of the library since the message response inbox is opaque to those systems.  I'll also note that this is quite a clever way to manage retrieving the `opId` out of the subject of the inboxes, which (as you mentioned) could also be used to further optimize our response processing in the future.

Awesome work, thanks for the contribution!

natewoods-wf:

rmconsole-wf (approved):
+1 from RM

PR #1627, opened 2021-07-19
fix nats 2.2 for python 3 and go

danielharasymiw-wf:
### Story:
Nats 2.2+ can return empty messages with a status header set.

### Acceptance Criteria:

### Design Notes:
Tried to follow pattern that was done with Java: https://github.com/Workiva/frugal/pull/1608/
### How To Test:
Run examples against nats 2.2

#### Reviewers:
@Workiva/service-platform


aviary2-wf:
## Security Insights
<details>
</details>

## Action Items
---

lib/go/nats_transport.go
```diff
@@ -84,7 +84,11 @@ func (f *fNatsTransport) Open() error {
 
 // handler receives a NATS message and executes the frame
 func (f *fNatsTransport) handler(msg *nats.Msg) {
-	if err := f.fBaseTransport.ExecuteFrame(msg.Data); err != nil {
+	_, hasStatus := msg.Header["Status"]
+	if len(msg.Data) == 0 && hasStatus {
```
brettkail-wk, line 88: For Java, I chose to completely ignore all messages with `Status` header since that seemed more conservative ([code](https://github.com/Workiva/frugal/blob/27c9a064ff72d6ef5fb6e0dcfa5f540e6aa1b535/lib/java/src/main/java/com/workiva/frugal/transport/FNatsTransport.java#L151-L159)).  If you think it's better to only exclude length 0, then that's fine with me, but I think it would be good to make the implementations consistent.

lib/python/frugal/aio/transport/async_transport.py
```diff
@@ -75,21 +89,36 @@ async def flush(self, payload):
...
         """
         Complete the future associated with the data frame.
 
         Args:
             frame: The response frame
         """
-        if not frame:
+        if not message.data:
+            logger.debug("Received empty message")
+            if message.subject:
```
brettkail-wk, line 105: It seems like this should be checking the status header rather than assuming that all messages with an empty payload are service-not-available?

I am surprised that service-not-available was implemented at all for Python but not Go.  I would have thought that Go would be a higher priority for us.

lib/go/nats_transport.go
danielharasymiw-wf: 👍  I think keeping these as consistent as possible between each other is a good goal.

lib/python/frugal/aio/transport/async_transport.py
danielharasymiw-wf: The nats py library doesn't support headers yet :(

brettkail-wk: Then I think it's a mistake to try to implement service-not-available.  It seems reasonable to drop empty messages since they can never be valid and with a comment saying they can occur for service-not-available.

danielharasymiw-wf:
breaking this up into multiple PRs
Go: https://github.com/Workiva/frugal/pull/1635

PR #1651, opened 2021-08-25
Give consumers the ability to filter out services, service methods, structs, and scopes from generation

joshprzybyszewski-wf:
Disclaimer: this PR is (currently) 958 lines of test code; only 530 lines of diff.

## Problem

Sometimes a consumer of frugal may only want a subset of an IDL to be generated.

#### Client Example
A client application needs to talk to a server for a single feature. The client will only ever use one of the frugal endpoints (a single service method) on a specific IDL `service` which has 100 methods defined on it. Today, the client application needs to generate code for all 100 methods, compile the generated code, and ship it to the browser. This generates a lot of unnecessary code that bloats file size unnecessarily.

It is currently impossible to only generate a subset of a frugal IDL.

## Solution

Let's provide consumers of frugal a way to filter out portions of an idl that they don't need or want to generate. This will only ever pare down generated code, and will not add new generated code.

### Considerations
1. Maintain backwards compatibility
   - This change needs to be opt-in, and must not break existing usage of frugal
2. Give consumers the option to "forcibly include" and to "exclude"
   - If something is defined in the `includes`, then it _will_ be there.
   - Otherwise, if something is in the `excludes`, then it won't be in the generated output.
3. Provide the option for golang and dartlang
   - I'm only interested in these languages, and don't consider myself an expert in the others. By no means am I precluding the other languages from adding this feature themselves.
   - Someday :tm:, we could add support for other languages
4. Provide options to filter by service methods, by structs, and by scopes.
   - These are the biggest offenders I've seen internally, so they're the ones I want to tackle first.
   - Eventually we could target constants, type defs, and other generated frugal code.

### Suggested Implementation
In order to meet all of these considerations, I suggest the following:
1. Add a new "Option" to the frugal command on a per-language level
   - This option specifies the path to a YAML file.
   - If left empty, it defaults to `frugal_filter.yaml` in the current directory.
   - we do so for golang here: https://github.com/Workiva/frugal/blob/e57f96ad10ca8b69dd4d5a192bd6cd6592275246/compiler/generator/generator.go#L59 and dartlang here: https://github.com/Workiva/frugal/blob/e57f96ad10ca8b69dd4d5a192bd6cd6592275246/compiler/generator/generator.go#L80
2. After we have parsed a given IDL and before we start generating code, we apply a filter to the parsed IDL.
   - This is done perfectly as the first thing in `func (o *programGenerator) Generate(frugal *parser.Frugal, outputDir string) error {`
   - We can mutate the state of the `*parser.Frugal` struct, and remove the things we don't want.
   - We opt in to the filter here: https://github.com/Workiva/frugal/blob/e57f96ad10ca8b69dd4d5a192bd6cd6592275246/compiler/generator/generator.go#L181-L184
   - We apply the filter here: https://github.com/Workiva/frugal/blob/e57f96ad10ca8b69dd4d5a192bd6cd6592275246/compiler/generator/base.go#L33-L41
3. Filtering is applied to the `*parser.Frugal` struct based on the options in the YAML file.
   - The YAML is read into a struct in memory.
   - The struct will know what to include/exclude.
   - We iterate through scopes, services, and structs to find out what the caller wants removed/kept.
   - Mutate the `*parser.Frugal` struct. Subsequent usage is left un-touched and should operate as if the definition wasn't in the IDL.
   - This all happens in the new `filter` package, and the root of it can be found here: https://github.com/Workiva/frugal/blob/e57f96ad10ca8b69dd4d5a192bd6cd6592275246/compiler/generator/filter/apply.go#L7-L23

## Potential Concerns
1. create more dependency tracking work (because generated artifacts are exported and people may be importing things that are now changing. :soon: we will be encouraging consumers to generate their own code from other repo's IDLs rather than publishing generated code from IDLs in their own repo)
2. Maintenance of a frugal_filter.yaml file in the wild can be gnarly, especially when it silently excludes "a new thing" that you expect to be there.
3. Is there a non-frugal specific solution that can achieve the same results? i.e. splitting source IDL files, pre-processing IDL files (like `#ifdef`), maintaining two separate `.frugal` files with full/slim features.

I'll address each concern below.

### Will this complicate dependency tracking?
No. 

Right now people rely on generated artifacts that are exported. This is a bad practice, and we should not encourage repos following this pattern to opt-in to this change. This new option will not affect these repos unless someone manually decides to use it (and they shouldn't use it without understanding it).

The target audience of this change is client repos (like the one I represent) that follow the good practice. The client repo will fetch the server's IDL files, then generate its own code that it does not publish for anyone else to use. By using this PR, my team's client-side repo sees a reduction in the generated code of >60% per file (because we don't need every method/scope/struct defined in the server IDL).

🤦  I responded to 2. and 3. last week, but somewhere in history, github has dropped my responses. tl;dr measuring the pros & cons of a change like this vs. other ideas, I think this path has the most pros with the least cons.

## Acceptance Criteria:

### How To Test:
Using this branch, you are able to create a yaml file specification to exclude/include portions of the IDL that you are generating (see example in internal Workiva repo).

### My Test Results:
In a private repo, I have seen a significant decrease in the size of output code (many packages by >60%).

#### Reviewers:
@Workiva/service-platform


joshprzybyszewski-wf:
By using the newly exported functions (from https://github.com/Workiva/frugal/pull/1660), consumers will be able to effectively wrap the frugal command and do whatever filtering they deem necessary.

I will close this PR because, at this time, frugal is not encouraging new features to be added since it is going through the process of containment.

PR #1659, opened 2021-08-30
Give consumers the option to omit server generation for a parsed service (in golang generation)

joshprzybyszewski-wf:
### Story:
Some clients use the golang generator to generate services, but they're only going to be clients of the IDLs; not servers. Therefore, when we generate the "server" code, we are creating unnecessary artifacts.

Let's give consumers to option to "omit generating the server code" for all services. This should only ever be used by consumers of frugal who are generating code for golang clients.

### Acceptance Criteria:

### Design Notes:
I've added a single config option to the golang generator which reads it right before it attempts to generate the server code for a service.

### How To Test:
1. Retains backwards compatibility
   - Using the frugal command (without the `omit_server_service_generation` option) outputs the exact same files before and after this PR.
2. Gives an opt-in option to generate less code:
   - While generating go code, use the `omit_server_service_generation` option and see that everything is exactly the same, except now you don't have a server in the generated output.

### My Test Results:
In a private Workiva repo, we see a significant reduction in the amount of unnecessary go code being generated.

#### Reviewers:
@Workiva/service-platform


aviary3-wk:
## Security Insights
<details>
</details>

## Action Items
---

benechols-wf (approved):

natewoods-wf (approved):
I'm aware this has been discussed at length in private, but just so it is known publically, this could also be achieved by removing the `service` sections of `*.frugal` files before generating bindings.  That said, this seems... fine.

natewoods-wf:

rmconsole-wf (approved):
+1 from RM

joshprzybyszewski-wf:
@natewoods-wf 
> this could also be achieved by removing the service sections of *.frugal files before generating bindings.

Perhaps I'm misunderstanding, but I don't think that's quite true. As a consumer, if I were to remove the `service` definition in the .frugal files when I generate golang frugal code, then I would no longer be generating the `FServiceNameClient` struct. There are cases where I want to still be a client without also being a server.

natewoods-wf:
Ah, correct.  Apologies, my allergy meds are messing with my head.  Nice work!

PR #1760, opened 2022-01-14
PP-5479 Fix Dart Frugal `hashCode` with Collections

timothysteward-wk:
### Story:
I noticed we were using `DeepCollectionEquality()` for the `==` operator in dart, but not for the `hashCode` functinon. We should also use `DeepCollectionEquality()` for the `hashCode` so we get the same `hashCode` when two objects have lists, maps, or sets with the same contents.

Before we were doing this to get the hashcode for a list:
```
value = (value * 31) ^ this.field;
```

We should be doing this:
```
value = (value * 31) ^ DeepCollectionEquality().hash(this.events);
```

We already use DeepCollectionEquality() for the == operator. The problem is that hashCode and == are not consistent with each other because == treats deeply equal collections as equal while hashCode does not. In dartlang == and hashCode are supposed to match.

> Hash codes must be the same for objects that are equal to each other according to operator ==. The hash code of an object should only change if the object changes in a way that affects equality. There are no further requirements for the hash codes. They need not be consistent between executions of the same program and there are no distribution guarantees.

From: https://api.flutter.dev/flutter/dart-core/Object/hashCode.html

### Acceptance Criteria:

### Design Notes:
TODO

### How To Test:
TODO

### My Test Results:
TODO

#### Reviewers:
@Workiva/service-platform


aviary3-wk:
## Security Insights
<details>
<summary>(4) Vulnerable direct dependencies were detected</summary>
</details>

## Action Items
---

compiler/generator/dartlang/generator.go
```diff
@@ -1308,15 +1308,23 @@ func (g *Generator) generateHashCode(s *parser.Struct) string {
...
 	return contents
 }
 
+func (g *Generator) generateFieldHashCode(field *parser.Field) string {
+	fieldName := toFieldName(field.Name)
+	if !g.isDartCollection(field.Type) {
+		return fmt.Sprintf("value = (value * 31) ^ this.%s.hashCode;\n", fieldName)
+	}
+
+	return fmt.Sprintf("value = (value * 31) ^ DeepCollectionEquality().hash(this.%s);\n", fieldName)
```
patkujawa-wf, line 1325: Where is `DeepCollectionEquality` imported from? `import 'package:collection/collection.dart';`? Does that import already exist?

tsteward: Yes, this already exists. Already using it for the `==` operator overload. This PR makes the `==` dart operator overload and `hashCode` consistent, when they previously weren't.

timothysteward-wk: Yes, this already exists. Already using it for the `==` operator overload. This PR makes the `==` dart operator overload and `hashCode` consistent, when they previously weren't.

patkujawa-wf (approved):

compiler/generator/dartlang/generator.go
alanknight-wk: This seems reasonable enough. The DeepCollectionEquality hash combination function seems at least as good as what we're using here, and I believe it's compatible with the one that's built in to Dart 2.14+. The current one seems like it would start to fail once we hit 11 fields and our hash value becomes a Javascript float.

It might be reasonable, and save some code size if the hash implementation was just
`    DeepCollectionEquality.hash([myField1, myField2, ...])`

josephnaberhaus-wk:
Working on QA review

robbecker-wf:
QA+1 Existing tests all pass with the changes to hashcode

PR #1884, opened 2022-05-28
Add support for java thread pool workers

adamdockter-wk:
Add support for java thread pool workers to help prevent messages from backing up in nats

### Story:
Adding support for java concurrent executors on the Nats subscribers.  This support is added to help prevent issues with nats topics getting backed up and causing exceptions when a handler execution is slow.

The default uses a single processing thread for all messages, same as before this change.

### Acceptance Criteria:

#### Reviewers:
@Workiva/service-platform

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
@@ -15,6 +15,8 @@ package com.workiva.frugal.transport;
 
 import com.workiva.frugal.exception.TTransportExceptionType;
 import com.workiva.frugal.protocol.FAsyncCallback;
+import com.workiva.frugal.util.DirectExecutor;
+
 import io.nats.client.Connection;
 
 import io.nats.client.Connection.Status;
@@ -27,6 +29,9 @@ import org.slf4j.LoggerFactory;
 
 import static com.workiva.frugal.transport.FNatsTransport.FRUGAL_PREFIX;
 
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+
 /**
  * FNatsSubscriberTransport implements FSubscriberTransport by using NATS as the pub/sub message broker.
  * Messages are limited to 1MB in size.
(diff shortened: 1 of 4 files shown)


aviary-wf:
## Security Insights
<details>
<summary>(6) Vulnerable direct dependencies were detected</summary>
<li>1 vulns in <code>gopkg.in/yaml.v3 < 3.0.0</code> via <code>go.sum</code></li>
</details>

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java
```diff
@@ -83,13 +95,39 @@ public Factory(Connection conn, String queue) {
...
+         * message are handled one at a time on a single thread.
+         * @param count The number of concurrent workers for consuming messages. Must be >= 1.
+         * @return The existing Factory
+         * @throws IllegalArgumentException if count < 1
+         */
+        public Factory withWorkerCount(int count) {
+            if( count < 1) {
+                throw new IllegalArgumentException("Worker count must be 1 or greater, but was " + count);
+            }
+            this.workerPool = Executors.newFixedThreadPool(count);
```
brettkail-wk, line 116: It seems kind of a "hazard" that there's no way for clients to control the lifecycle of these threads when they probably should be shutting down the executor (and there's no pre-existing `close` method in this class to add it automatically).  I wonder if we should only add `withWorkerPool` in this code?

That said, the way we've solved this in the past is to have the subscriber submit to the thread pool itself.  Does this need to be solved in Frugal, or could it be solved in the subscriber instead?

brettkail-wk: Discussed with @kevinfrommelt-wf, and it sounds like a repo-specific solution wasn't given much consideration before adding that Frugal Go API.  If we're only adding a pool method, then I'm not sure a `.withWorkerPool(pool)` is significantly nicer than `.subscribeX(r -> pool.execute(r))`, so I lean towards not adding new API, but if you feel strongly that it's worth it, then I guess go ahead.

PR #1891, opened 2022-06-01
Updgrade thrift to 0.14.2 to avoid cve

adamdockter-wk:
### Story:
There were some change to some of the Thrift Interfaces to add some additional method. 

* FProtocol is mostly a straight pass through to TProtocol
* TFramedTransport is a straight pass through to TTransport
* TMemoryOutputBuffer on the other hand was rebased to TEndpointTransport to support the default implementations of the missing abstract methods on TTransport, this also caused some exception handle propagation to FNatsServer

### Acceptance Criteria:

#### Reviewers:
@Workiva/service-platform

### DIFF
--- a/lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
+++ b/lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
@@ -122,6 +122,11 @@ public class FProtocol extends TProtocol {
     }
 
     @Override
+    public int getMinSerializedSize(byte type) throws TException {
+        return wrapped.getMinSerializedSize(type);
+    }
+
+    @Override
     public void writeMessageBegin(TMessage tMessage) throws TException {
         wrapped.writeMessageBegin(tMessage);
     }
--- a/lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
@@ -30,10 +30,9 @@ import java.io.ByteArrayOutputStream;
  */
 public class TMemoryOutputBuffer extends TTransport {
 
-    private ByteArrayOutputStream buffer;
-    private final int limit;
+    private final ByteArrayOutputStream buffer;
     private final byte[] emptyFrameSize = new byte[4];
-    private static final TConfiguration configuration = new TConfiguration();
+    private final TConfiguration configuration;
 
     /**
      * Create an TMemoryOutputBuffer with no buffer size limit.
@@ -45,12 +44,13 @@ public class TMemoryOutputBuffer extends TTransport {
     /**
      * Create an TMemoryOutputBuffer with a buffer size limit.
      *
-     * @param size the size limit of the buffer. Note: If <code>size</code> is non-positive,
-     *             no limit will be enforced on the buffer.
+     * @param maxMessageSize the size limit of the buffer. Note: If <code>size</code> is
+     *             non-positive, no limit will be enforced on the buffer.
      */
-    public TMemoryOutputBuffer(int size) {
-        buffer = new ByteArrayOutputStream(size);
-        limit = size;
+    public TMemoryOutputBuffer(int maxMessageSize) {
+        buffer = new ByteArrayOutputStream();
+        configuration = new TConfiguration(maxMessageSize,
+            TConfiguration.DEFAULT_MAX_FRAME_SIZE, TConfiguration.DEFAULT_RECURSION_DEPTH);
         init();
     }
 
@@ -84,10 +84,11 @@ public class TMemoryOutputBuffer extends TTransport {
 
     @Override
     public void write(byte[] buf, int off, int len) throws TTransportException {
-        if (limit > 0 && buffer.size() + len > limit) {
+        int maxMessageSize = configuration.getMaxMessageSize();
+        if (maxMessageSize > 0 && buffer.size() + len > maxMessageSize) {
             reset();
             throw new TTransportException(
-                    TTransportExceptionType.REQUEST_TOO_LARGE, String.format("Buffer size reached (%d)", limit));
+                    TTransportExceptionType.REQUEST_TOO_LARGE, String.format("Buffer size reached (%d)", maxMessageSize));
         }
         buffer.write(buf, off, len);
     }
--- a/lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
@@ -28,16 +28,17 @@ import java.io.ByteArrayOutputStream;
  * The size of this buffer is optionally limited. If limited, writes which cause the buffer to exceed
  * its size limit throw an TTransportException with code TTransportExceptionType.REQUEST_TOO_LARGE.
  */
-public class TMemoryOutputBuffer extends TEndpointTransport {
+public class TMemoryOutputBuffer extends TTransport {
 
     private ByteArrayOutputStream buffer;
     private final int limit;
     private final byte[] emptyFrameSize = new byte[4];
+    private static final TConfiguration configuration = new TConfiguration();
 
     /**
      * Create an TMemoryOutputBuffer with no buffer size limit.
      */
-    public TMemoryOutputBuffer() throws TTransportException {
+    public TMemoryOutputBuffer() {
         this(0);
     }
 
@@ -47,11 +48,9 @@ public class TMemoryOutputBuffer extends TEndpointTransport {
      * @param size the size limit of the buffer. Note: If <code>size</code> is non-positive,
      *             no limit will be enforced on the buffer.
      */
-    public TMemoryOutputBuffer(int size) throws TTransportException {
-        super(new TConfiguration());
+    public TMemoryOutputBuffer(int size) {
         buffer = new ByteArrayOutputStream(size);
         limit = size;
-        updateKnownMessageSize(size);
         init();
     }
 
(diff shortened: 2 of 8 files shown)


aviary-wf:
## Security Insights
<details>
<summary>(5) Vulnerable direct dependencies were detected</summary>
</details>

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/protocol/FProtocol.java
```diff
@@ -121,6 +121,11 @@ public void readResponseHeader(FContext context) throws TException {
         context.addResponseHeaders(headers);
     }
 
+    @Override
```
amitjain-wk, line 124: Are all the below changes triggered because of version bump ?

adamdockter-wk: yes there were new abstract method in the new version that caused these extra changes

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
```diff
@@ -44,9 +47,11 @@ public TMemoryOutputBuffer() {
      * @param size the size limit of the buffer. Note: If <code>size</code> is non-positive,
      *             no limit will be enforced on the buffer.
      */
-    public TMemoryOutputBuffer(int size) {
-        buffer = new ByteArrayOutputStream();
+    public TMemoryOutputBuffer(int size) throws TTransportException {
+        super(new TConfiguration());
+        buffer = new ByteArrayOutputStream(size);
         limit = size;
+        updateKnownMessageSize(size);
```
brettkail-wk, line 54: The `throws TTransportException` is a breaking API change, and it seems avoidable.  If nothing else, suggest wrapping the `updateKnownMessageSize` in ` try`/`catch (...) { throw new IllegalStateException }`.

But I don't really understand the intent of this `updateKnownMessageSize` call since it doesn't seem necessary.  It doesn't look like there are any callers in Thrift, so perhaps this should continue to `extends TTransport` and just implement the new `TTransport` methods with `throw new UnsupportedOperationException`?

adamdockter-wk: I feel it was a bit of an oversight on Thrift's part to make those abstract. They are used by the built in TMemoryBuffer which implements both read and write buffer functionality and those three new abstract methods are all to support read limits, since we are make an out only memory buffer they are pointless to our implementation. 

I will go the approach of wrapping the exception some sort of runtime exception that doesn't require a checked exception signature.

brettkail-wk: > we are make an out only memory buffer they are pointless

Agreed, so then why not implement them to throw exceptions rather than extending `TEndpointTransport` since they should never be called?

adamdockter-wk: I made the change, but I am not really sure it makes it any more or less of a breaking change.  I am not sure why it was considered a breaking change before.  I guess because people may have been creating their own TMemoryOutputBuffer?  The usage of it in NatServer was already handling the exception.  TMemoryInputBuffer which is not ours will still throw the checked exception so I still need to make a change in there to handle it.  Part of me thinks the symmetry between the exceptions for input and output are better.  But I am fine either way.

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
```diff
@@ -30,6 +33,7 @@ public class TMemoryOutputBuffer extends TTransport {
     private ByteArrayOutputStream buffer;
     private final int limit;
     private final byte[] emptyFrameSize = new byte[4];
+    private static final TConfiguration configuration = new TConfiguration();
```
brettkail-wk, line 36: Out of curiosity, do you know how/if this is actually used?  It seems like this would more accurately be `new TConfiguration(size, DEFAULT_MAX_FRAME_SIZE, DEFAULT_RECURSION_DEPTH)` (using the size param in constructor).  Perhaps it's not actually used, or perhaps it would require generator changes to be used?

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
brettkail-wk: > I guess because people may have been creating their own TMemoryOutputBuffer?

Right.  I have historically found [this wiki page](https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_interfaces_-_API_methods) to be clearest.

> TMemoryInputBuffer which is not ours will still throw the checked exception so I still need to make a change in there to handle it.  Part of me thinks the symmetry between the exceptions for input and output are better. But I am fine either way.

We can't control Thrift API breaks, but we can control our own.  The new APIs seem intended to limit input (attacker controlled), so the general approach in the PR now looks good to me.  (I guess a single "transport" interface merging both input/output is awkward, and I guess this is an extension of that awkwardness.)

adamdockter-wk: Sounds good, we can move forward with it this way, thanks for the wiki link

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
```diff
@@ -45,7 +49,7 @@ public TMemoryOutputBuffer() {
      *             no limit will be enforced on the buffer.
      */
     public TMemoryOutputBuffer(int size) {
-        buffer = new ByteArrayOutputStream();
+        buffer = new ByteArrayOutputStream(size);
```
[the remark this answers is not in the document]
adamdockter-wk: This was an oversight on my part. I noticed the input buffer was passing the size through the the input stream.  So I thought it would be appropriate to do the same here. I can undo this

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
adamdockter-wk: Let me do some more checking on this.  I had just notice other transports would return a default instance (instead of null) of the configuration when it wasn't supplied.  But I just had a thought that this is also probably a mutable object and I probably should not return a single static instance.

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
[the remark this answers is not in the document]
brettkail-wk: Undoing makes sense.  Out of curiosity, do you have a link to the input buffer allocating a `ByteArrayOutputStream` like that?  For input, I would guess it means "this is the exact inbound request size", so it would make some sense since the buffer is allocated in full (vs this output API where it means "this is the maximum outbound request size").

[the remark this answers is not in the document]
adamdockter-wk:
```java
/**
 * Memory buffer-based implementation of the TTransport interface.
 */
public class TMemoryBuffer extends TEndpointTransport {
  /**
   * Create a TMemoryBuffer with an initial buffer size of <i>size</i>. The
   * internal buffer will grow as necessary to accommodate the size of the data
   * being written to it.
   *
   * @param size the initial size of the buffer
   */
  public TMemoryBuffer(int size) throws TTransportException {
    super(new TConfiguration());
    arr_ = new TByteArrayOutputStream(size);
    updateKnownMessageSize(size);
  }
```
https://github.com/apache/thrift/blob/master/lib/java/src/main/java/org/apache/thrift/transport/TMemoryBuffer.java
But I think I didn't clearly read the docs and one is an initial size and the other is a limit, I renamed the parameter to limit to be more clear.

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
adamdockter-wk: Good suggestion, now that I realize the size param is a limit I opted to remove the limit field completely and just use the state stored in the configuration object

brettkail-wk (approved):

adamdockter-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

lib/java/src/main/java/com/workiva/frugal/transport/TMemoryOutputBuffer.java
```diff
@@ -41,12 +44,13 @@ public TMemoryOutputBuffer() {
...
-     *             no limit will be enforced on the buffer.
+     * @param maxMessageSize the size limit of the buffer. Note: If <code>size</code> is
+     *             non-positive, no limit will be enforced on the buffer.
      */
-    public TMemoryOutputBuffer(int size) {
+    public TMemoryOutputBuffer(int maxMessageSize) {
         buffer = new ByteArrayOutputStream();
-        limit = size;
+        configuration = new TConfiguration(maxMessageSize,
+            TConfiguration.DEFAULT_MAX_FRAME_SIZE, TConfiguration.DEFAULT_RECURSION_DEPTH);
```
patkujawa-wf, line 53: Should this be max message size, not max frame size?

adamdockter-wk: This constructor 
```
public TConfiguration(int maxMessageSize, int maxFrameSize, int recursionLimit) {
```

patkujawa-wf: 🤦🏻 thanks, I skipped right over that first arg

PR #1949, opened 2022-07-27
MEX-395 config size

adamdockter-wk:
### Story:
We updated thrift to 0.14.2 which introduced a TConfiguration for managing limits.  In the first pass for this upgrade we missed some places where we were holding on to limits via integers.  This was causing some limits to not be respected. 

As part of this change:

- improved some of the unit tests around checking limits
- changed input transport constructors to pass TConfiguration
- introduced a TConfigurationBuilder that treats 0 as a MAX integer limit

### Acceptance Criteria:

#### Reviewers:
@Workiva/service-platform

### DIFF
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>3.16.0.TEST-SNAPSHOT</version>
+    <version>3.16.0</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
--- a/lib/java/pom.xml
+++ b/lib/java/pom.xml
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>3.16.0</version>
+    <version>3.16.0.TEST-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
@@ -59,7 +59,7 @@ public class FHttpTransport extends FTransport {
     // Immutable
     private final CloseableHttpClient httpClient;
     private final String url;
-    private final TConfiguration responseConfig;
+    private final int responseSizeLimit;
     private final FHttpTransportHeaders requestHeaders;
 
     private FHttpTransport(CloseableHttpClient httpClient, String url, int requestSizeLimit, int responseSizeLimit,
@@ -68,7 +68,7 @@ public class FHttpTransport extends FTransport {
         this.requestSizeLimit = requestSizeLimit;
         this.httpClient = httpClient;
         this.url = url;
-        this.responseConfig = TConfigurationBuilder.custom().setMaxMessageSize(responseSizeLimit).build();
+        this.responseSizeLimit = responseSizeLimit;
         this.requestHeaders = requestHeaders;
     }
 
@@ -211,6 +211,8 @@ public class FHttpTransport extends FTransport {
 
         byte[] response = makeRequest(context, payload);
 
+        TConfiguration responseConfig =
+            TConfigurationBuilder.custom().setMaxMessageSize(responseSizeLimit).build();
         return response == null ? null : new TMemoryInputTransport(responseConfig, response);
     }
 
@@ -277,14 +279,14 @@ public class FHttpTransport extends FTransport {
 
         request.setHeader("accept", "application/x-frugal");
         request.setHeader("content-transfer-encoding", "base64");
-        if (responseConfig.getMaxMessageSize() > 0) {
-            request.setHeader("x-frugal-payload-limit", Integer.toString(responseConfig.getMaxMessageSize()));
+        if (responseSizeLimit > 0) {
+            request.setHeader("x-frugal-payload-limit", Integer.toString(responseSizeLimit));
         }
         request.setEntity(requestEntity);
         request.setConfig(RequestConfig.custom()
                 .setConnectTimeout((int) context.getTimeout())
                 .setSocketTimeout((int) context.getTimeout())
-                .build());
+            .build());
 
         // Make request
         CloseableHttpResponse response;
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
@@ -212,7 +212,7 @@ public class FHttpTransport extends FTransport {
         byte[] response = makeRequest(context, payload);
 
         TTransport responseTransport = null;
-        if(response != null) {
+        if (response != null) {
             TConfiguration responseConfig =
                 TConfigurationBuilder.custom().setMaxMessageSize(responseSizeLimit).build();
             responseTransport = new TMemoryInputTransport(responseConfig, response);
--- a/lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
+++ b/lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
@@ -28,6 +28,7 @@ import org.apache.http.entity.AbstractHttpEntity;
 import org.apache.http.entity.ContentType;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.EntityUtils;
+import org.apache.thrift.TConfiguration;
 import org.apache.thrift.transport.TMemoryInputTransport;
 import org.apache.thrift.transport.TTransport;
 import org.apache.thrift.transport.TTransportException;
@@ -58,7 +59,7 @@ public class FHttpTransport extends FTransport {
     // Immutable
     private final CloseableHttpClient httpClient;
     private final String url;
-    private final int responseSizeLimit;
+    private final TConfiguration responseConfig;
     private final FHttpTransportHeaders requestHeaders;
 
     private FHttpTransport(CloseableHttpClient httpClient, String url, int requestSizeLimit, int responseSizeLimit,
@@ -66,8 +67,8 @@ public class FHttpTransport extends FTransport {
         super();
         this.httpClient = httpClient;
         this.url = url;
-        this.requestSizeLimit = requestSizeLimit;
-        this.responseSizeLimit = responseSizeLimit;
+        this.requestConfig = TConfigurationBuilder.custom().setMaxMessageSize(requestSizeLimit).build();
+        this.responseConfig = TConfigurationBuilder.custom().setMaxMessageSize(responseSizeLimit).build();
         this.requestHeaders = requestHeaders;
     }
 
@@ -210,7 +211,7 @@ public class FHttpTransport extends FTransport {
 
         byte[] response = makeRequest(context, payload);
 
-        return response == null ? null : new TMemoryInputTransport(response);
+        return response == null ? null : new TMemoryInputTransport(responseConfig, response);
     }
 
     private static class Base64EncodingEntity extends AbstractHttpEntity {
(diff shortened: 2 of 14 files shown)


aviary2-wf:
## Security Insights
<details>
<summary>(6) Vulnerable direct dependencies were detected</summary>
<li>1 vulns in <code>aiohttp</code> via <code>lib/python/requirements_dev_asyncio.txt</code> with no fix reported by GitHub</li>
</details>

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/transport/FTransport.java
```diff
@@ -37,7 +39,7 @@ public abstract class FTransport {
     private volatile FTransportClosedCallback monitor;
     private boolean isOpen;
 
-    protected int requestSizeLimit;
+    protected TConfiguration requestConfig = TConfigurationBuilder.DEFAULT;
```
brettkail-wk, line 42: This seems to be a breaking API change, which seems unwise.  Do we need to make this change?  It seems like we should instead keep the fields as-is and just ensure that we use the existing config vars creating the `TMemoryInputStream`.

adamdockter-wk: Is the worry that someone inheriting from this would have to update to work with the requestConfig instead of the int?

brettkail-wk: Yes.

adamdockter-wk: I definitely could change it, I don't like the change as much as this one.  I suppose if I make a breaking change like this I would need to bump at least a minor version instead of a patch?

brettkail-wk: IMO, always avoid API breaks when possible, and it seems trivial to avoid here (just create a fresh `TConfiguration` in `request`.)

(`TConfiguration` is surprisingly mutable.  If you strongly prefer to break API, then suggest creating a fresh instance so that `requestConfig.setMaxMessageSize(x)` doesn't change the _global_ default.)

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
```diff
@@ -276,8 +277,8 @@ private byte[] makeRequest(FContext context, byte[] requestPayload) throws TTran
 
         request.setHeader("accept", "application/x-frugal");
         request.setHeader("content-transfer-encoding", "base64");
-        if (responseSizeLimit > 0) {
-            request.setHeader("x-frugal-payload-limit", Integer.toString(responseSizeLimit));
+        if (responseConfig.getMaxMessageSize() > 0) {
+            request.setHeader("x-frugal-payload-limit", Integer.toString(responseConfig.getMaxMessageSize()));
```
ryanjackson-wf, line 281: I think that the terminology may be confusing me, but this appears to be using the response configuration in establishing a request. Maybe a comment here would help if this is intentional.

lib/java/src/main/java/com/workiva/frugal/transport/FTransport.java
ryanjackson-wf: Given that this issue is blocking at least some services from upgrading frugal/m-sdk as part of the token migration, we might leave this be and create a follow-up ticket to re-examine this.

lib/java/src/main/java/com/workiva/frugal/transport/TConfigurationBuilder.java
```diff
@@ -0,0 +1,21 @@
...
+public class TConfigurationBuilder {
+  public static final TConfiguration DEFAULT = TConfiguration.DEFAULT;
+  private final TConfiguration.Builder builder = TConfiguration.custom();
+
+  public static TConfigurationBuilder custom() {
+    return new TConfigurationBuilder();
+  }
+
+  public TConfigurationBuilder setMaxMessageSize(int maxMessageSize) {
+    builder.setMaxMessageSize(maxMessageSize > 0 ? maxMessageSize : Integer.MAX_VALUE);
```
ryanjackson-wf, line 14: Should we leave the "else" of this ternary as zero? That would preserve the intention and normal usage of the thrift library. It would also help to preserve the existing behavior.

ryanjackson-wf: Actually, it would allow us to drop the ternary completely unless we wanted to check for values less than zero.

brettkail-wk: This is effectively an adapter from "frugal-style max" (where <=0 means "unlimited") to "thrift-style max" (where <=0 means "must be empty").  Renaming the helper to `TConfigurationUtils.ofOptionalMaxRequestSize(int max)` might be clearer (or just embed it in the few places where it matters).

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
```diff
@@ -210,7 +211,9 @@ public TTransport request(FContext context, byte[] payload) throws TTransportExc
 
         byte[] response = makeRequest(context, payload);
 
-        return response == null ? null : new TMemoryInputTransport(response);
+        TConfiguration responseConfig =
+            TConfigurationBuilder.custom().setMaxMessageSize(responseSizeLimit).build();
+        return response == null ? null : new TMemoryInputTransport(responseConfig, response);
```
brettkail-wk, line 216: I guess no reason to create config for null response?
```
if (response == null) {
  return null;
}

TConfiguration responseConfig = ...
return new ...
```

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
adamdockter-wk: I believe this is adding a header to the request so the receiving service knows what the expected response size limit it.

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
```diff
@@ -283,7 +286,7 @@ private byte[] makeRequest(FContext context, byte[] requestPayload) throws TTran
         request.setConfig(RequestConfig.custom()
                 .setConnectTimeout((int) context.getTimeout())
                 .setSocketTimeout((int) context.getTimeout())
-                .build());
+            .build());
```
brettkail-wk, line 289: This seems worse or at least inconsistent.

brettkail-wk (approved):

lib/java/pom.xml
```diff
@@ -3,7 +3,7 @@
 
     <groupId>com.workiva</groupId>
     <artifactId>frugal</artifactId>
-    <version>3.16.0</version>
+    <version>3.16.0.TEST-SNAPSHOT</version>
```
brettkail-wk, line 6: (Approved, but will need to revert this.)

lib/java/src/main/java/com/workiva/frugal/transport/TConfigurationBuilder.java
ryanjackson-wf: 🤔 I see what you mean. I had been going back and forth between repos and conflated/confused them.

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
ryanjackson-wf: I kind of figured that was what was happening, but I thought I'd ask to be sure. Thank you.

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
```diff
@@ -64,9 +65,9 @@ public class FHttpTransport extends FTransport {
     private FHttpTransport(CloseableHttpClient httpClient, String url, int requestSizeLimit, int responseSizeLimit,
             FHttpTransportHeaders requestHeaders) {
         super();
+        this.requestSizeLimit = requestSizeLimit;
```
ryanjackson-wf, line 68: [nit] You might as well leave this alongside the response size limit.

lib/java/src/main/java/com/workiva/frugal/transport/TConfigurationBuilder.java
adamdockter-wk: @ryanjackson-wf do you know if there is documentation or code that shows zero is treated and infinite?  I was looking at [TEndpointTransport.java](https://github.com/apache/thrift/blob/7a5f29d48d6b99f14190d9d7bff0e63170965087/lib/java/src/main/java/org/apache/thrift/transport/TEndpointTransport.java#L60) and it appears to me that zero in Thrift would reject any payload.
```java
    protected void resetConsumedMessageSize(long newSize) throws TTransportException {
        // full reset
        if (newSize < 0)
        {
            knownMessageSize = getMaxMessageSize();
            remainingMessageSize = getMaxMessageSize();
            return;
        }

        // update only: message size can shrink, but not grow
        if (newSize > knownMessageSize)
            throw new TTransportException(TTransportException.END_OF_FILE, "MaxMessageSize reached");

        knownMessageSize = newSize;
        remainingMessageSize = newSize;
    }
```

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
adamdockter-wk: I will refactor

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
adamdockter-wk: will do

lib/java/src/main/java/com/workiva/frugal/transport/TConfigurationBuilder.java
ryanjackson-wf: Brett was correct. I had spent so much time switching back and forth between frugal and thrift that I confused their default behavior. His suggestion (and your implementation) of the ternary is correct.

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
adamdockter-wk: non-intentional I will revert

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
```diff
@@ -210,7 +211,13 @@ public TTransport request(FContext context, byte[] payload) throws TTransportExc
 
         byte[] response = makeRequest(context, payload);
 
-        return response == null ? null : new TMemoryInputTransport(response);
+        TTransport responseTransport = null;
+        if(response != null) {
```
ryanjackson-wf, line 215:
```suggestion
        if (response != null) {
```

lib/java/src/main/java/com/workiva/frugal/transport/FTransport.java
```diff
@@ -17,6 +17,8 @@
 import com.workiva.frugal.exception.TTransportExceptionType;
 import com.workiva.frugal.transport.monitor.FTransportMonitor;
 import com.workiva.frugal.transport.monitor.MonitorRunner;
+
+import org.apache.thrift.TConfiguration;
```
ryanjackson-wf, line 21: Is this import still needed?

adamdockter-wk: nope, removed

lib/java/src/main/java/com/workiva/frugal/transport/FHttpTransport.java
adamdockter-wk: fixed

ryanjackson-wf (approved):

adamdockter-wk:
QA +1 - Larry Lynn Tested via wTranslate https://workiva.slack.com/archives/C03R1QBD3D4/p1658966445630699?thread_ts=1658944132.371699&cid=C03R1QBD3D4

adamdockter-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

Issue #1536, opened 2021-03-27
go: failed generating correct code


sakura-rip:
when running below command,
`frugal -gen=go:async hoge.frugal`

The generated file will raise `Unresolved reference 'ctx'` Error
![image](https://user-images.githubusercontent.com/56917703/112717352-6a9f6d00-8f2f-11eb-8a63-3f5bb39d9d35.png)

https://github.com/Workiva/frugal/blob/daacef7dce45875fdf620095f03dc06a920664b3/compiler/generator/golang/generator.go#L1979

This location is wrong.

PR #2206, opened 2023-05-24
Fix async go compiler to use correct argument

kevinsookocheff-wf:
### Story:

Compiling the music store example using the async go generator results in an `undeclared name: ctx` error:

```
func (f *FStoreClient) BuyAlbumAsync(fctx frugal.FContext, asin string, acct string) (r <-chan *Album, err <-chan error) {
	errC := make(chan error, 1)
	resultC := make(chan *Album, 1)
	go func() {
		result, err := f.BuyAlbum(ctx, asin, acct)  // undeclared name: ctx
```

Addresses issue https://github.com/Workiva/frugal/issues/1536

### Acceptance Criteria:

#### Reviewers:
@Workiva/service-platform

### DIFF
--- a/compiler/generator/golang/generator.go
+++ b/compiler/generator/golang/generator.go
@@ -1975,7 +1975,7 @@ func (g *Generator) generateHandlerArgs(method *parser.Method) string {
 	return args
 }
 func (g *Generator) generateCallArgs(method *parser.Method) string {
-	args := "ctx"
+	args := "fctx"
 	for _, arg := range method.Arguments {
 		args += ", " + strings.ToLower(arg.Name)
 	}
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ unit_tests_py3.xml
 
 # Cross-compiled code
 pkg/
+
+# macos
+.DS_Store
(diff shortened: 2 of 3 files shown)


compiler/generator/golang/generator.go
```diff
@@ -1975,7 +1975,7 @@ func (g *Generator) generateHandlerArgs(method *parser.Method) string {
 	return args
 }
 func (g *Generator) generateCallArgs(method *parser.Method) string {
```
brettkail-wk, line 1977: `generateAsyncCallArgs` might be a better method name.  It took me a moment to figure out why everything isn't broken :-).

brettkail-wk (approved):

compiler/generator/golang/generator.go
kevinsookocheff-wf: Yeah there are so many places this could all be cleaned up :). I was going to start working on these things in 10% time as I'm able.

kevinsookocheff-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #2209, opened 2023-05-25
dart: Add FGeneratedArgsResultBase

brettkail-wk:
### Story:
Generated code for Dart is larger than desired.

This is the first commit of [this work-in-progress branch](https://github.com/Workiva/frugal/compare/develop...brettkail-wk:frugal:dart_smaller).  The final commit there states:
```
The _args/_result structs are only used internally by the generated
client methods, so those structs don't need the full generality of
generated structs, so:
1.  use use_null_for_unset, which produces smaller code
2.  omit methods that not used by the generated client methods
```
The intent of this PR is to merge+release this base class to simplify deployment.

### Acceptance Criteria:
- [x] Code has been tested and results documented
- [x] Unit tests have been updated
- ~~[ ] Updates to documentation if necessary~~
- ~~[ ] Verify and document changes to any other Messaging components~~
- [x] Pull request made against the 'develop' branch, not master

### Design Notes:
 builds on this commit by modifying 

### How To Test:
CI only (new/unused API only).

### My Test Results:
I manually tested the work-in-progress against a large internal application that currently produces ~270KLOC of Dart, and  the branch reduces that to ~220KLOC.

#### Reviewers:
@Workiva/service-platform
@robbecker-wf @kevinsookocheff-wf

PR #2221, opened 2023-06-02
dart: Generate smaller code

brettkail-wk:
### Story:
Use `FGeneratedArgsResultBase` added in #2209 to reduce the size of generated dart code

### Acceptance Criteria:
- ~~[ ] Updates to documentation if necessary~~
- ~~[ ] Verify and document changes to any other Messaging components~~

### Design Notes:
Commits are intended to be reviewed independently.  See commit messages for details.  Ultimately, the diff is +255/-177 on the generator code, and the rest of the changes are to the `compiler/testdata/expected`.

In particular, the removal of the write/read commits sufficiently confuses the diff algorithm to make the test changes unreadable (the `_args`/`_result` class lines no longer delineate the changes).  It's also easier to see the impact of each individual change by reviewing commit-by-commit.

### My Test Results:
I ran this branch against the full test suite of a large internal application that currently generates ~270KLOC of Dart, and this branch reduces it to ~205KLOC, which reduces the dart2js output by ~340KB.

#### Reviewers:
@Workiva/service-platform
@robbecker-wf @kevinsookocheff-wf


aviary2-wf:
## Security Insights

## Action Items
---

brettkail-wk:
Rebased to resolve two issues:
1. Replaced commit "dart: Fix default values for null_for_unset" with commit "dart: Add TODO to field initialization" commit.  I mistakenly thought that default values weren't being initialized (from code inspection, not testing), but the fields were being initialized when declared, so the old commit was re-initializing fields (so all testing passed), which was already the case for non-`use_null_for_unset`.  Add a comment to make the suboptimal generated code clearer.
4. Added commit "dart: Fix use_null_for_unset default requiredness".  While inspecting the changes of the "dart: Don't generate isSet/unset for _args/_result", I noticed that `isSet` calls were being added to `write`, and that's because `use_null_for_unset` was wrongly checking `isSet` for fields of "primitive" type that were using default requiredness.

compiler/generator/base.go
```diff
@@ -80,8 +80,13 @@ func (b *BaseGenerator) GetElem() string {
 	return s
 }
 
-func (b *BaseGenerator) GetServiceMethodTypes(service *parser.Service) []*parser.Struct {
-	structs := []*parser.Struct{}
+type Struct struct {
+	*parser.Struct
+	Args bool
```
adamdockter-wk, line 85: As discussed in the code review, it would be nice to have a comment next to this boolean to describe what Args == false is since it is not apparent it would mean result struct.

brettkail-wk: Renamed to `ServiceMethodStruct`, and added struct- and field-level doc.

compiler/testdata/expected/dart.nullunset/actual_base/f_thing.dart
```diff
@@ -48,19 +48,11 @@ class thing implements thrift.TBase {
   setFieldValue(int fieldID, Object value) {
     switch (fieldID) {
       case AN_ID:
-        if (value == null) {
-          unsetAn_id();
-        } else {
-          this.an_id = value as int;
-        }
+        this.an_id = value as dynamic;
```
danielszopa-wk, line 51: This was the chunk that confused me yesterday.

> ... with Null safety enabled, a variable declared as dynamic can be null, while a variable of type Object cannot be null.

My guess is that when we want to support null saftey we can write this like...

```dart
  setFieldValue(int fieldID, dynamic value) {
    switch (fieldID) {
      case AN_ID:
        this.an_id = value;
```

That being said this looks good for now 👍

compiler/testdata/expected/dart.int64/actual_base/f_base_foo_service.dart
```diff
@@ -191,30 +135,6 @@ class basePing_result extends frugal.FGeneratedArgsResultBase {
...
-  int get hashCode {
-    var value = 17;
-    return value;
-  }
-
-  basePing_result clone() {
-    return basePing_result();
-  }
-
   validate() {
```
danielszopa-wk, line 109: I'm not familiar with the full semantics of `validate()`, but it seems like we could maybe remove this for args & results as well?

This does look like it might be a bit more complex as you'd have to remove the `validate()` calls from `read()` and `write()` as well. [Here's an example](https://sourcegraph.plat.workiva.net/github.com/Workiva/frugal/-/blob/examples/dart/gen-dart/v1_music/lib/src/f_track.dart?L407-412) where it seems to be generated with an actual implementation.

danielszopa-wk (approved):

compiler/testdata/expected/dart.nullunset/actual_base/f_thing.dart
brettkail-wk: The issue is that it's [implementing this method](https://github.com/Workiva/thrift-dart/blob/3059908dc516048f8aabb24434db2aff7ad1e773/lib/src/t_base.dart#L36), and I don't think we can change that without breaking everything, so we'll probably need the `as dynamic` forever.

@robbecker-wf I would have expected `setFieldValue(int, Object?)` so that the pre-existing `== null` checks continue to work.  Was that an oversight with the null-safety changes?  Similarly, I assume `dynamic?` is correct for the type for `getFieldValue`, but it's not specified at all; is that the default?

adamdockter-wk (approved):

charliestrawn-wf (approved):

compiler/testdata/expected/dart.nullunset/actual_base/f_thing.dart
brettkail-wk: Discussed offline with Rob, and he agreed that `setFieldValue(int, dynamic)` would break everything and that `setFieldValue(int, Object?)` is the correct signature for null-safety.  There will need to be some changes in the future to add some kind of `--gen dart:null_safe` option.

brettkail-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #2282, opened 2023-08-22
FEDX-145 : Update frugal compiler to generate null safe code

alishadalal-wk:
Update frugal compiler to generate null safe code


aviary-wf:
## Security Insights

## Action Items
---

compiler/generator/dartlang/generator.go
```diff
@@ -35,7 +35,7 @@ const (
 	defaultOutputDir    = "gen-dart"
```
brettkail-wk, line 35: (Arbitrary line.)

Are there more changes coming?  The `?` needs to be added all over to the generated code (struct fields/params, service params/results, etc.).

compiler/generator/dartlang/generator.go
```diff
@@ -198,13 +198,13 @@ func (g *Generator) addToPubspec(dir string) error {
 	pubFilePath := filepath.Join(dir, "pubspec.yaml")
 
 	deps := map[interface{}]interface{}{
-		"collection": "^1.14.12",
+		"collection": "^1.15.0",
 		"logging":    "^1.0.0",
 		"thrift": dep{
 			Hosted:  hostedDep{Name: "thrift", URL: "https://pub.workiva.org"},
 			Version: "^0.0.14",
```
robbecker-wf, line 205:
```suggestion
			Version: "^0.0.15",
```

alishadalal-wk: If we update this, getting error in skynet: 

```
Because every version of frugal_test from path depends on thrift ^0.0.15 which depends on http ^0.13.0, every version of frugal_test from path requires http ^0.13.0.
2023-08-23T12:09:32.986248532Z So, because frugal_test_client depends on both http ^0.12.0 and frugal_test from path, version solving failed.
```

robbecker-wf: http can and should be updated to ^0.13.3

compiler/generator/dartlang/generator.go
```diff
@@ -1177,11 +1203,35 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, kind structKind, f
...
+                "eventsDefault": true,
+                "success": true,
+                "listfield": true,
+                "list2": true,
+                "list3": true,
+                "list4": true,
+
+            }
+            if validNames[fName] {
+                contents += fmt.Sprintf(tab+ind+"%s%s?.add(%s);\n", thisPrefix, fName, valElem)
```
brettkail-wk, line 1231: Hardcoding names here doesn't make sense here either.  The problem you're hitting is actually first=false vs true: if we're accessing a local-var, then it's non-null, but if we're accessing `this.` then it is.

One proper solution would be to append `?` only if `!first`.  I think it would be better (more idiomatic Dart) to always generate the local var as non-nullable, and then assign to the field at the end.  To do that, I would push the `prefix` var down into the `if` branches.  For primitive+struct, that will be a small amount of duplicate code to setup the `prefix` for.  For the container branch, remove the `prefix` var altogether, define the local var as non-nullable, then populate, then assign to `this.` at the end.  That way, you don't need `?.add`, `![]`, etc.

compiler/generator/dartlang/generator.go
```diff
@@ -1325,26 +1405,80 @@ func (g *Generator) generateWriteFieldRec(field *parser.Field, first bool, ind s
...
+
+        // Using a map to add ! for the nullable types
+        validNames := map[string]bool{
+            "eventMapDefault": true,
+            "eventMap": true,
+            "req": true,
+        }
+        for key := range validNames {
+            if strings.Contains(fName, key) {
+                fName = strings.Replace(fName,key,key+"!",1)
```
brettkail-wk, line 1417: For this write code path, you want to check `first` and add a `!` only if `first`.  It might make sense to declare a single `thisName = "this." + fName + "!"` instead of `thisPrefix`, I'm not sure.

For the container branch, it seems most natural to output a declaration of a non-nullable local var (similar to read), and then use that.

compiler/generator/dartlang/generator.go
```diff
@@ -955,9 +955,35 @@ func (g *Generator) generateFieldMethods(s *parser.Struct, kind structKind) stri
...
 		if !g.useNullForUnset(kind) {
 			contents += g.generateFieldComment(field, tab)
-			contents += fmt.Sprintf(tab+"%s get %s => this._%s;\n\n", dartType, fName, fName)
+
+			nullableDartTypes := map[string]struct{}{
+            	"int":                              {},
+            	"String":                           {},
+            	"bool":                             {},
+            	"Set<t_variety.Event>":             {},
+            	"Map<fixnum.Int64, t_variety.Event>": {},
```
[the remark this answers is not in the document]
alishadalal-wk: Removed this specific condition & updated to `dartType+"?"` for both get/set

compiler/generator/dartlang/generator.go
```diff
@@ -1177,11 +1203,35 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, kind structKind, f
 		case "list":
 			contents += fmt.Sprintf(ind+"thrift.TList %s = iprot.readListBegin();\n", containerElem)
 			contents += ignoreDeprecationWarningIfNeeded(ind, field.Annotations)
-			contents += fmt.Sprintf(ind+"%s%s = %s;\n", prefix, fName, dartType)
+
+            // Convert to list literal
+            var dartType2 = strings.Replace(dartType,"List","",1);
```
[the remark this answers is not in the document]
alishadalal-wk: Ok. I have found this change at one place only. So, should we create a separate PR for the suggested changes?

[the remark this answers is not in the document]
brettkail-wk: I would have thought that `Set`/`Map` needed the same, but I guess it's just warnings instead?  Anyway, if it's isolated to this one place and you'd rather do that refactor as part of this PR, I guess that's up to you.

compiler/generator/dartlang/generator.go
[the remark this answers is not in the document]
alishadalal-wk: Updated the same

compiler/generator/dartlang/generator.go
alishadalal-wk: Added the logic to generate local var as non-nullable & then assign to `this.` at the end. Removed `?.add, ![]` also

compiler/generator/dartlang/generator.go
alishadalal-wk: Tried the logic to add a `!` only if `first` but this does not works as expected (i.e. add `!` where required like `eventMapDefault[elem61]` as it should be updated to `eventMapDefault![elem61]` in _compiler/testdata/expected/dart/variety/f_event_wrapper.dart_ ). 

Is there any specific condition for this?

compiler/generator/dartlang/generator.go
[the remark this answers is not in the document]
brettkail-wk: Right, I don't like the `strings.Replace` post-processing of `getDartTypeFromThriftType`.  This code only wants the generic type parameters, not the full type with `List`, so factor out a method that only returns the `<t_variety.Events>`, which I would name `getDartGenericTypeArgsFromThriftType` or similar.  Use it in the `getDartTypeFromThriftType` method and also this call-site here.

compiler/generator/dartlang/generator.go
```diff
@@ -433,7 +433,7 @@ func (g *Generator) GenerateConstantsContents(constants []*parser.Constant) erro
 			immutableKeyword = "const"
 		}
 		contents += fmt.Sprintf(tab+"static %s %s %s = %s;\n",
-			immutableKeyword, g.getDartTypeFromThriftType(constant.Type), constant.Name, value)
+			immutableKeyword, g.getDartTypeFromThriftType(constant.Type)+"?", constant.Name, value)
```
brettkail-wk, line 436: (Arbitrary line.)

@robbecker-wf Is it safe to add these `?` unconditionally, or will that break projects that aren't ready to move to null-safe?  i.e., do we need to add a `--gen dart:null_safety` option or similar so that they're only added for projects that are ready to opt in?

robbecker-wf: 💡 Good point. I think we might need a switch to turn on generating nullsafe code like you suggest. MOST places just generate a standalone dart library, but there are some that generate Dart code that is included in a larger Dart package and that would be a mixed null safety situation.
@brettkail-wk How do we add something like that? Have 2 copies of the generator (original and non-nullable) ? 
Having one generator with `if` conditions sprinkled everywhere seems messy.

brettkail-wk: > MOST places just generate a standalone dart library

At least in RRPL, we always use `library_prefix`, so the generated pubspec.yaml isn't used, so the corresponding pubspec.yaml presumably needs to be updated at the same time ([example](https://github.com/Workiva/xbrl2-server-api/blob/master/api-dart/resources/pubspec.yaml)).

> How do we add something like that? Have 2 copies of the generator (original and non-nullable) ?
Having one generator with if conditions sprinkled everywhere seems messy.

I agree `if` everywhere is messy, but having two copies of the generator sounds even worse.  I would suggest:
- where possible, update generated code to work regardless of null safety (e.g., [this comment](https://github.com/Workiva/frugal/pull/2282#discussion_r1307614762))
- add `Nullable string` (used as `g.Nullable`) that is initialized to `""` (default) or `"?"` (if `null_safety` is enabled).  That way, the "if" is condensed to one spot in the code.
- if necessary, add a `NonNull string` initialized to `""` or `"!"`
- add `if` for remaining, which is hopefully few.

compiler/generator/dartlang/generator.go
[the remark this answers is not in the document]
alishadalal-wk: Okay, I got it. So, I have removed `strings.Replace`; added function **getDartGenericTypeArgsFromThriftType** and updated the code as required. Also updated it for `set/map` for null safety.

compiler/generator/dartlang/generator.go
```diff
@@ -1728,16 +1874,16 @@ func (g *Generator) GeneratePublisher(file *os.File, scope *parser.Scope) error
 	publishers += fmt.Sprintf("class %s {\n", publisherClassname)
 	publishers += tab + "frugal.FPublisherTransport transport;\n"
 	publishers += tab + "frugal.FProtocolFactory protocolFactory;\n"
-	publishers += tab + "Map<String, frugal.FMethod> _methods;\n"
+	publishers += tab + "Map<String, frugal.FMethod>? _methods = {};\n"
```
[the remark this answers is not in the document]
alishadalal-wk: I have updated it.

compiler/generator/dartlang/generator.go
```diff
@@ -2098,7 +2244,25 @@ func (g *Generator) generateClientMethod(service *parser.Service, method *parser
...
+			"sayAgain":              true,
+			"blah":                  true,
+			"param_modifiers":       true,
+			"getThing":              true,
+		}
+
+		if nonNullableFNames[nameLower] {
+			contents += tabtabtab + "return result.success!;\n"
+		} else {
+			contents += tabtabtab + "return result.success;\n"
```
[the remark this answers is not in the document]
alishadalal-wk: Removed above if-else condition and updated the following code to make return value nullable:

```
func (g *Generator) generateReturnArg(method *parser.Method) string {
	if method.ReturnType == nil {
		return ""
	}
	return fmt.Sprintf("<%s>", g.getDartTypeFromThriftType(method.ReturnType)+ "?")
}
```

But it will add `?` to all.

compiler/generator/dartlang/generator.go
```diff
@@ -1516,7 +1638,31 @@ func (g *Generator) generateClone(s *parser.Struct) string {
 	for _, field := range s.Fields {
 		fieldName := toFieldName(field.Name)
 		contents += ignoreDeprecationWarningIfNeeded(tabtab, field.Annotations)
-		contents += fmt.Sprintf(tabtab+"%s %s,\n", g.getDartTypeFromThriftType(field.Type), fieldName)
+
+		// Using a map to add ? for the nullable types
+		nullableNames := map[string]bool{
```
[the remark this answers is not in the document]
alishadalal-wk: Removed the same.

compiler/generator/dartlang/generator.go
```diff
@@ -2199,6 +2363,32 @@ func (g *Generator) isDartCollection(t *parser.Type) bool {
...
 }
 
+func (g *Generator) getDartGenericTypeArgsFromThriftType(t *parser.Type) string {
+	if t == nil {
+		return ""
+	}
+	underlyingType := g.Frugal.UnderlyingType(t)
+
+	if g.Frugal.IsEnum(underlyingType) {
+		return ""
```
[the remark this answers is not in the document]
alishadalal-wk: Updated the same.

compiler/generator/dartlang/generator.go
```diff
@@ -1316,6 +1359,15 @@ func (g *Generator) generateWriteFieldRec(field *parser.Field, first bool, ind s
 			panic("unknown thrift type: " + underlyingType.Name)
 		}
 
+		nullableFNames := map[string]string{
+			"req[elem68]":      "req![elem68]",
```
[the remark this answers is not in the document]
alishadalal-wk: Its been created dynamically such as `fmt.Sprintf("%s[%s]", fName, keyElem)`. So when we check the `fName` value in this block; we get combined string of `req[elem68]`. As we need to add `!` in some of these, so added this logic.

However, removed these hardcoding and updated the code.

compiler/generator/dartlang/generator.go
```diff
@@ -1325,26 +1377,80 @@ func (g *Generator) generateWriteFieldRec(field *parser.Field, first bool, ind s
 			contents += fmt.Sprintf(tabtab+ind+"oprot.writeI32(%s%s);\n", thisPrefix, fName)
 		}
 	} else if g.Frugal.IsStruct(underlyingType) {
+
+		//Using a map to add ! for the nullable types
+		validNames := map[string]bool{
+			"eventMapDefault": true,
```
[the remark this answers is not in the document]
alishadalal-wk: Removed this code & added ! using the condition of `first` 

```
nullCheckOperator := ""
if first {
	nullCheckOperator = "!"
}
...case list:
...
contents += fmt.Sprintf(tabtab+ind+"oprot.writeListBegin(thrift.TList(%s, %s%s%s.length));\n", valEnumType, thisPrefix, fName, nullCheckOperator)
...
contents += fmt.Sprintf(tabtab+ind+"for(var %s in %s%s%s) {\n", valElem, thisPrefix, fName, nullCheckOperator)

```
Defining local variable is not working here as we still need to add `!` as the some variables in class are nullable.

compiler/generator/dartlang/generator.go
```diff
@@ -1171,43 +1171,86 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, kind structKind, f
...
-		dartType := g.getDartTypeFromThriftType(underlyingType)
-
+		dartType := g.getDartGenericTypeArgsFromThriftType(underlyingType) // Get generic type
 		switch underlyingType.Name {
 		case "list":
 			contents += fmt.Sprintf(ind+"thrift.TList %s = iprot.readListBegin();\n", containerElem)
 			contents += ignoreDeprecationWarningIfNeeded(ind, field.Annotations)
-			contents += fmt.Sprintf(ind+"%s%s = %s();\n", prefix, fName, dartType)
+			localListVar := "tempList"
+			if prefix == "this." {
```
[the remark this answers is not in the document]
alishadalal-wk: Updated the logic as follows as thisPrefix will already have the value `this.` when `first` is **true** which will cause incorrect value of listVar = `this..<fName>`
```
initializer := "{}"
if underlyingType.Name == "list" {
  initializer := "[]"
}
...
...
localListVar := "tempList"
if first {
contents += fmt.Sprintf(ind+"var %s = %s%s;\n", localListVar, dartType, initializer)
} else {
contents += fmt.Sprintf(ind+"%s%s = %s%s;\n", prefix, fName, dartType,initializer)
}
```

compiler/generator/dartlang/generator.go
alishadalal-wk: Removed ?  from here as its used to define const vars. So we can remove it.

compiler/generator/dartlang/generator.go
```diff
@@ -1171,43 +1171,88 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, kind structKind, f
...
+		if underlyingType.Name == "list" {
+			initializer = "[]"
+		}
 		switch underlyingType.Name {
 		case "list":
 			contents += fmt.Sprintf(ind+"thrift.TList %s = iprot.readListBegin();\n", containerElem)
 			contents += ignoreDeprecationWarningIfNeeded(ind, field.Annotations)
-			contents += fmt.Sprintf(ind+"%s%s = %s();\n", prefix, fName, dartType)
+			localListVar := "tempList"
+			if first {
```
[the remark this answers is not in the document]
alishadalal-wk: Optimized & updated the code

compiler/generator/dartlang/generator.go
```diff
@@ -1171,43 +1171,88 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, kind structKind, f
...
+				contents += fmt.Sprintf(ind+"%s%s = %s%s;\n", prefix, fName, dartType, initializer)
+			}
 			contents += fmt.Sprintf(ind+"for(int %s = 0; %s < %s.length; ++%s) {\n", counterElem, counterElem, containerElem, counterElem)
 			contents += valContents
 			contents += ignoreDeprecationWarningIfNeeded(tab+ind, field.Annotations)
-			contents += fmt.Sprintf(tab+ind+"%s%s.add(%s);\n", thisPrefix, fName, valElem)
+			if first {
+				contents += fmt.Sprintf(tab+ind+"%s.add(%s);\n", localListVar, valElem)
+			} else {
+				contents += fmt.Sprintf(tab+ind+"%s%s.add(%s);\n", thisPrefix, fName, valElem)
```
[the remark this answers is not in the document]
alishadalal-wk: Updated the same.

compiler/generator/dartlang/generator.go
```diff
@@ -1325,37 +1370,46 @@ func (g *Generator) generateWriteFieldRec(field *parser.Field, first bool, ind s
...
+		} else {
+			contents += fmt.Sprintf(tabtab+ind+"%s%s!.write(oprot);\n", thisPrefix, fName)
+		}
+
 	} else if underlyingType.IsContainer() {
 		valEnumType := g.getEnumFromThriftType(underlyingType.ValueType)
-
+		nullCheckOperator := ""
+		if first {
+			nullCheckOperator = "!"
```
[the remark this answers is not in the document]
alishadalal-wk: Done. Updated the code as below:

```
localVar := fName
if first {
localVar = "temp"
contents += fmt.Sprintf(tabtab+ind+"final %s = this.%s!;\n", localVar, fName)
}
```

compiler/generator/dartlang/generator.go
```diff
@@ -1171,45 +1171,47 @@ func (g *Generator) generateReadFieldRec(field *parser.Field, kind structKind, f
...
-			contents += ignoreDeprecationWarningIfNeeded(ind, field.Annotations)
-			contents += fmt.Sprintf(ind+"%s%s = %s();\n", prefix, fName, dartType)
-			contents += fmt.Sprintf(ind+"for(int %s = 0; %s < %s.length; ++%s) {\n",
-				counterElem, counterElem, containerElem, counterElem)
-			contents += valContents
-			contents += fmt.Sprintf(tab+ind+"%s%s.add(%s);\n", thisPrefix, fName, valElem)
-			contents += ind + "}\n"
-			contents += ind + "iprot.readSetEnd();\n"
+			if !first {
+				localVar = thisPrefix + fName
```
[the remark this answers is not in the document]
alishadalal-wk: Ok. updated it as follows as else part needs **prefix** and not **thisPrefix**.
```
var localVar string
 if first {
   localVar = "temp" + underlyingNameWithCapitalType
   contents += fmt.Sprintf(ind+"var %s = %s%s;\n", localVar, dartType, initializer)
} else {
   localVar = thisPrefix + fName
   contents += fmt.Sprintf(ind+"%s%s = %s%s;\n", prefix, fName, dartType, initializer)
}
```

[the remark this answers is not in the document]
brettkail-wk: I've lost track of what we're doing.  At a high-level, we want to unconditionally declare a local var regardless of whether it's the first or recursive invocation, and then if it's the first invocation, we want to write the local var to the field at the end.  In the snippet you posted:
- We're defining the local var with `var` for first invocation but with a type (`prefix`) for the recursive case.  That seems unnecessary/confusing.
- In the `else`, if we're defining a local var, but we're not assigning `localVar` to use it.

Can it be simplified to this?
```
localVar := fName
if first {
  localVar := g.GetElem()
}
contents += fmt.Sprintf(ind+"final %s = %s%s;\n", localVar, dartType, initializer)
```
This should remove the need for `prefix`/`thisPrefix` altogether because we're always accessing a local var and we can remove the need for the type by using plain `final`.

(In the PR, it appears the `if !first { localVar = thisPrefix + fName }` is still present in each `case`, which can be removed.)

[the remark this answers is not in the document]
alishadalal-wk: `localVar` defined in the if-else condition code is being used in the each `case` and `if !first { localVar = thisPrefix + fName }` is already removed. **Code link -** [https://github.com/Workiva/frugal/blob/874aa7f1d07e7b080e783e365baaabb2fa24e4e5/compiler/generator/dartlang/generator.go#L1195](https://github.com/Workiva/frugal/blob/874aa7f1d07e7b080e783e365baaabb2fa24e4e5/compiler/generator/dartlang/generator.go#L1195)

Yes, we can simplify the if-else block as you suggested. So updated the code as per the same - [Link](https://github.com/Workiva/frugal/blob/c5cfe5e6a273d0d3996c7235301a668c163ff28b/compiler/generator/dartlang/generator.go#L1182)

 (Also removed thisPrefix from this function as its not required)

compiler/generator/dartlang/generator.go
```diff
@@ -955,9 +955,9 @@ func (g *Generator) generateFieldMethods(s *parser.Struct, kind structKind) stri
 
 		if !g.useNullForUnset(kind) {
 			contents += g.generateFieldComment(field, tab)
-			contents += fmt.Sprintf(tab+"%s get %s => this._%s;\n\n", dartType, fName, fName)
+			contents += fmt.Sprintf(tab+"%s? get %s => this._%s;\n\n", dartType, fName, fName)
```
brettkail-wk, line 958: (Arbitrary line.)

The new codegen broadly looks ok to me, so I think [this thread](https://github.com/Workiva/frugal/pull/2282#discussion_r1322103991) is all that remains.

robbecker-wf:
Notes on edit, generate, refresh cycle

In frugal
`make unit-cli-copy && go install`
// describe what these do
make unit-cli - just runs the generator and compares the output to the files in the expected directory
make unit-cli-copy - runs the unit tests and copies the ouput into the expected directory
go install - installs the latest changes and puts the `frugal` command in your path so you can run it from anywhere

In licensing_frugal
`make package` the first time. it generates code and copies pubspec.yaml into gen/dart/licensing_frugal
`make generate` after each generator.go update. Regenerates the dart code only

What's left?
change the generator.go :
- testdata/expected/dart and dart.int64 and dart.nullunset are all unchanged when compared to "develop" branch
- testdata/expected/dart.nullsafe generates nullsafe code with sdk constrain 2.12 and ? and ! in the right places
- verify that licensing_frugal generates nullsafe code without null errors

compiler/generator/dartlang/generator.go
```diff
@@ -371,7 +401,11 @@ func (g *Generator) GenerateFile(name, outputDir string, fileType generator.File
 
 // GenerateDocStringComment generates the autogenerated notice.
 func (g *Generator) GenerateDocStringComment(file *os.File) error {
-	comment := fmt.Sprintf(
+	comment := ""
+	if addDartComment {
+		comment += fmt.Sprintf("%s\n", g.dartComment)
```
brettkail-wk, line 406: Out of curiosity, what does this comment do?

Why check `if addDartComment` if it's always true?

compiler/generator/dartlang/generator.go
```diff
@@ -1728,14 +1762,25 @@ func (g *Generator) GeneratePublisher(file *os.File, scope *parser.Scope) error
...
+
+	if g.genNullsafe {
+		publishers += tab + "Map<String, frugal.FMethod> _methods = {};\n"
+		publishers += fmt.Sprintf(tab+"%s(frugal.FScopeProvider provider, [List<frugal.Middleware>%s middleware]) :\n", publisherClassname, g.nullableOperator)
+		publishers += tabtab + "transport = provider.publisherTransportFactory.getTransport(),\n"
+		publishers += tabtab + "protocolFactory = provider.protocolFactory\n"
+		publishers += tab + "{\n"
+		publishers += tabtab + "var combined = middleware ?? [];\n"
+		publishers += tabtab + "combined.addAll(provider.middleware);\n"
+	} else {
```
brettkail-wk, line 1774: Why separate code paths?  That makes it hard to see how the code is different.  It seems like we can generate an initializer in both cases?

compiler/generator/dartlang/generator.go
```diff
@@ -1991,14 +2041,19 @@ func (g *Generator) generateClient(service *parser.Service) string {
 			clientClassname, servTitle)
 	}
 	contents += fmt.Sprintf(tab+"static final logging.Logger _frugalLog = logging.Logger('%s');\n", servTitle)
-	contents += tab + "Map<String, frugal.FMethod> _methods;\n\n"
+
+	if g.genNullsafe {
+		contents += tab + "Map<String, frugal.FMethod> _methods = {};\n\n"
+	} else {
+		contents += tab + "Map<String, frugal.FMethod> _methods;\n\n"
```
brettkail-wk, line 2048: Why this difference?

compiler/generator/dartlang/generator.go
```diff
@@ -1966,11 +2011,16 @@ func (g *Generator) getServiceExtendsName(service *parser.Service) string {
...
+	if g.genNullsafe {
+		lateKeyword = g.lateOperator + " "
+	}
 	contents := ""
 
 	// Generate client factory
-	contents += fmt.Sprintf("%s %sFactory(frugal.FServiceProvider provider, {List<frugal.Middleware> middleware}) =>\n",
-		clientClassname, lowercaseFirstCharacter(clientClassname))
+	contents += fmt.Sprintf("%s %sFactory(frugal.FServiceProvider provider, {List<frugal.Middleware>%s middleware}) =>\n",
+		clientClassname, lowercaseFirstCharacter(clientClassname), g.nullableOperator)
```
brettkail-wk, line 2023: Why is `late` required?  Why can't we use initializer lists?

compiler/generator/dartlang/generator.go
alishadalal-wk: The first block is added for generating code in case of null safety **(as analyzer was giving error - Non-nullable instance field `<varName>` must be initialized.)** , while the second block is intended to preserve the existing code(in non-nullsafe folder) as it is.

compiler/generator/dartlang/generator.go
alishadalal-wk: It basically adds a dart comment at the top of the file - like following:
```
(For dart nullsafety code)
// @dart=2.12

(For dart non-nullsafety code)
// @dart=2.11
```

@robbecker-wf , shall we remove the condition - `if addDartComment` condition as it is redundant (because it's always true)?

compiler/generator/dartlang/generator.go
alishadalal-wk: same reason as above

compiler/generator/dartlang/generator.go
robbecker-wf: Adding these comments was helpful when developing since it forces treating a dart file as either nullable or nullsafe. What was happening was the dart files in the other expected data directories was being treated as nullsafe and showing errors in VS Code. So by adding 2.11 to those, it cleaned up the noise.

So, we could leave the flag in, set as false but add a comment that says, enable this when debugging Dart generated code.

compiler/generator/dartlang/generator.go
robbecker-wf: right, unless it's a real pain, I'd like to have the generator generate the exact same code as before (without the nullsafe cmd line arg). Then we can guarantee that we haven't broken anything for existing users (until we migrate them by adding the cmdline arg. And then we have the chance to test and verify more)

brettkail-wk: > I'd like to have the generator generate the exact same code as before

And for repo maintenance, I'd prefer we _didn't_ do that.  If the generated code needs to change slightly+harmlessly when `-gen dart:nullsafe` is unspecified, then we should do that because if in the future we need to make other changes, we don't want 2^numFeatures copies of code.

If we're nervous about making changes to the non-nullsafe path in this PR, then we should open a separate PR to make those changes first.

Anyway, I'll defer to the service platform team.

robbecker-wf: For context, I see the nullsafe arg as temporary. As soon as we regenerate all the Dart code with null safety, the plan would be to come back and remove the ability to generate old code. So, I'm willing to take on a little tech debt temporarily to have a safe rollout.

robbecker-wf: ETA for removing the old code path would be in the weeks, not months or years.

compiler/generator/dartlang/generator.go
alishadalal-wk: Yes, We can use initializer list. So removed the `late` keyword &  updated the code for the same. Also, the code is adjusted the preserve the old code as it is and apply changes to null safe code only.

Link - [https://github.com/Workiva/frugal/pull/2282/commits/ddbeb070ae040115efa802037d49eab0c7d5cba0#diff-ad3b0590e79afa039372580f9643848c1993c0bd173a220ab99ad9c6118ad236L2053](https://github.com/Workiva/frugal/pull/2282/commits/ddbeb070ae040115efa802037d49eab0c7d5cba0#diff-ad3b0590e79afa039372580f9643848c1993c0bd173a220ab99ad9c6118ad236L2053)

compiler/generator/dartlang/generator.go
alishadalal-wk: Updated the code for the same.

robbecker-wf:
I made a licensing_frugal PR with both nullable and nullsafe modes and am happy with the output.
https://github.com/Workiva/licensing_frugal/pull/789
latest commit is nullsafe, this commit is nullable https://github.com/Workiva/licensing_frugal/pull/789/commits/1927c74b91257206881fd99282ed7a6d41e13ac2

robbecker-wf:
QA+1 CI passes and testing in other repos looks good in nullable and nullsafe code. The nullsafe changes are behind a command line argument so people won't immediate get the nullsafe generated code.

robbecker-wf:
+10 I'm going to go ahead with this as I'm confident in this at this point. @Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #2308, opened 2023-09-18
Attempts at fixing up compiler

robbecker-wf:
Work In Progress

- Added a command line flag
- Added test data for nullsafe
- made some fixes to generation

Things that are still weird and maybe there's even more than this but
- We changed the write method to use the temp localVar but that seems to have added other nullability issues like .length and map access that must be `!` 
- straight up fields don't seem to have the `?` on them when they should
- `List<frugal.Middleware> middleware` should have a `?` on the type but it doesn't get added
- The nullsafe test data doesn't generate a pubspec.yaml

I think most of that ^ has been resolved, but could still be weird. Also, would like to get the changes to generated code to only affect the nullsafe output and no changes to nullable code.


compiler/generator/dartlang/generator.go
```diff
@@ -372,7 +378,7 @@ func (g *Generator) GenerateFile(name, outputDir string, fileType generator.File
 // GenerateDocStringComment generates the autogenerated notice.
 func (g *Generator) GenerateDocStringComment(file *os.File) error {
 	comment := fmt.Sprintf(
-		"// Autogenerated by Frugal Compiler (%s)\n"+
+		"// Autogenerated by Frugal Compiler (nullsafe) (%s)\n"+
```
brettkail-wk, line 381: You probably don't want this (or only want to do this conditionally).

aviary-wf:
## Security Insights

## Action Items
---

compiler/generator/dartlang/generator.go
```diff
@@ -947,6 +956,10 @@ func (g *Generator) generateFieldComment(field *parser.Field, indent string) str
 func (g *Generator) generateFieldMethods(s *parser.Struct, kind structKind) string {
 	// Getters and setters for each field
 	contents := ""
+	nullable := ""
+	if g.nullsafe() {
+		nullable = "?"
```
brettkail-wk, line 961: Suggest a new `Nullable` field initialized in `NewGenerator` (or a new `nullable` method alongisde `nullsafe`) so that you don't have to do this in each method.

compiler/generator/generator.go
```diff
@@ -78,6 +78,7 @@ var Languages = LanguageOptions{
 		"use_int64":          "Use fixnum.Int64 for i64 fields",
 		"use_null_for_unset": "Use null for unset fields, which disallows null-but-set fields",
 		"use_vendor":         "Use specified import references for vendored includes and do not generate code for them",
+		"nullsafe":           "Generate nullsafe Dart code",
```
brettkail-wk, line 81: I still think it would be good to add an `expected/dart.nullsafe` directory + test.

compiler/generator/dartlang/generator.go
```diff
@@ -2113,7 +2142,11 @@ func (g *Generator) generateReturnArg(method *parser.Method) string {
 	if method.ReturnType == nil {
 		return ""
 	}
-	return fmt.Sprintf("<%s>", g.getDartTypeFromThriftType(method.ReturnType)+"?")
+	nullable := ""
+	if g.nullsafe() {
+		nullable = "?"
+	}
+	return fmt.Sprintf("<%s>", g.getDartTypeFromThriftType(method.ReturnType)+nullable)
```
brettkail-wk, line 2149: Use `%s` rather than `+` for nullable as you did in other cases?

compiler/generator/dartlang/generator.go
robbecker-wf: haha yeah that was to make sure the code was doing something at first. It should be removed.

compiler/generator/generator.go
robbecker-wf: I agree.

compiler/compiler_test.go
```diff
@@ -72,6 +72,43 @@ func TestValidDartFrugalCompiler(t *testing.T) {
 	suite.Run(t, options)
 }
 
+func TestValidDartNullsafe(t *testing.T) {
+	options := compiler.Options{
+		File:    frugalGenFile,
+		Gen:     "dart:nullsafe",
+		Out:     outputDir,
```
robbecker-wf, line 79: The other ones in this file append to outputDir to make it more unique, but that seems to cause problems where it doesn't get the nullsafe config correctly and doesn't generate the right null or nullable code. Which is weird.
@brettkail-wk ?

compiler/compiler_test.go
```diff
@@ -72,6 +72,43 @@ func TestValidDartFrugalCompiler(t *testing.T) {
...
+		{"dart.nullsafe/variety/f_testing_defaults.dart", "variety/lib/src/f_testing_defaults.dart"},
+		{"dart.nullsafe/variety/f_testing_unions.dart", "variety/lib/src/f_testing_unions.dart"},
+		{"dart.nullsafe/variety/f_health_condition.dart", "variety/lib/src/f_health_condition.dart"},
+		{"dart.nullsafe/variety/f_test_lowercase.dart", "variety/lib/src/f_test_lowercase.dart"},
+		{"dart.nullsafe/variety/f_foo_args.dart", "variety/lib/src/f_foo_args.dart"},
+		{"dart.nullsafe/variety/f_variety_constants.dart", "variety/lib/src/f_variety_constants.dart"},
+		{"dart.nullsafe/variety/f_events_scope.dart", "variety/lib/src/f_events_scope.dart"},
+		{"dart.nullsafe/variety/f_foo_service.dart", "variety/lib/src/f_foo_service.dart"},
+		{"dart.nullsafe/variety/variety.dart", "variety/lib/variety.dart"},
+		{"dart.nullsafe/variety/pubspec.yaml", "variety/pubspec.yaml"},
```
robbecker-wf, line 98: adding in these pubspec lines gets the files copied over to the expected test data location

alishadalal-wk (approved):

robbecker-wf:
merging into another PR branch

PR #2328, opened 2023-10-06
Simpler example

kevinsookocheff-wf:
@allyfinger-wk


aviary-wf:
## Security Insights

## Action Items
---

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.thrift.transport.TTransportException;
+import v1.music.PurchasingError;
+
+public class ExceptionHandlerServerMiddleware implements ServiceMiddleware {
+
+    @Override
+    public <T> InvocationHandler<T> apply(T next) {
+        return new InvocationHandler<T>(next) {
+            @Override
+            public Object invoke(Method method, Object receiver, Object[] args) throws Throwable {
```
kevinsookocheff-wf, line 18: When the server receives a request, it will hit this middleware (if we add the middleware to the server. See HttpServer.java).

The arguments `method`, `receiver`, and `args` are the underlying RPC call to make to the server (in this case the "buyAlbum" call).

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+                System.out.printf("==== CALLING %s.%s ====\n", method.getDeclaringClass().getName(), method.getName());
+
+                // Default error code is UNKNOWN
+                int errorCode = TApplicationException.UNKNOWN;
+                try {
+                    try {
+                        Object result = method.invoke(receiver, args);
```
kevinsookocheff-wf, line 25: Here we call the underlying implementation. This calls FStoreHandler.buyAlbum in our case. That method has been configured to throw and exception one out of three times.

kevinsookocheff-wf: `InvocationTargetException` is just Java's way of saying I tried to call method.invoke and that call threw an error. We can get the underlying exception thrown using `e.getTargetException()`.

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+                        return result;
+                    } catch (InvocationTargetException e) {
+                        // Throw underlying exception
+                        throw e.getTargetException();
+                    }
+                } catch (TApplicationException e) {
```
kevinsookocheff-wf, line 31: `TApplicationException` and `TTransportException` have defined error codes in the linked files. We can use `e.getType()` to return the int.

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+                    // See https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/exception/TApplicationExceptionType.java
+                    // and https://github.com/Workiva/messaging-sdk/blob/master/lib/java/sdk/src/main/java/com/workiva/messaging_sdk/exception/TApplicationExceptionType.java
+                    errorCode = e.getType();
+                } catch (TTransportException e) {
+                    // See https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/exception/TTransportExceptionType.java
+                    // and https://github.com/Workiva/messaging-sdk/blob/master/lib/java/sdk/src/main/java/com/workiva/messaging_sdk/exception/TTransportExceptionType.java
+                    errorCode = e.getType();
+                } catch (PurchasingError e) {
```
kevinsookocheff-wf, line 39: In a real implementation, this last `catch` would just be `Exception` and use `UNKNOWN` or maybe a new error code we create for SLOs called `UNDEFINED` or something.

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+                    // We can't really do this in a generic way because this is a user defined struct
+                    // Best we can do is a generic error code
+                    //
+                    // Maybe that is okay for MVP and if folks want better granularity they can update their code or use this middleware
+                    // as a base for their own (which isn't that uncommon)
+                    errorCode = TApplicationException.UNKNOWN;
```
kevinsookocheff-wf, line 45: We could maybe do something with reflection and then any user defined exception that has a `type` field would automatically get an error code (though that may conflict with existing codes).

```
e.getClass().getMethod("getType", new Class<?>[] {}).invoke(e);
```

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
brettkail-wk: I believe the try/catch is unnecessary because [this logic](https://github.com/Workiva/frugal/blob/3cf3c22ee15636de2cf2790eb5fd968e637429f6/lib/java/src/main/java/com/workiva/frugal/middleware/InvocationHandler.java#L39) already unwraps.

brettkail-wk: 🤦 But it's of course required for the outer `try`/`catch` to work.

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+                        Object result = method.invoke(receiver, args);
+                        return result;
+                    } catch (InvocationTargetException e) {
+                        // Throw underlying exception
+                        throw e.getTargetException();
+                    }
+                } catch (TApplicationException e) {
+                    // See https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/exception/TApplicationExceptionType.java
+                    // and https://github.com/Workiva/messaging-sdk/blob/master/lib/java/sdk/src/main/java/com/workiva/messaging_sdk/exception/TApplicationExceptionType.java
+                    errorCode = e.getType();
```
brettkail-wk, line 34: It seems a poor example to swallow the exception.  Perhaps rethrow and put the `errorCode` logging in the `finally`?

To be sure, server middleware won't see INTERNAL_ERROR for uncaught `RuntimeException` (like `NullPointerException`) because that translation doesn't happen until `FBaseProcessor`, so you have to catch those (or `Throwable`) explicitly.  I think updating `FStoreHandler` to periodically throw `NullPointerException` would show that.

I thought we had done work to also handle `Error` (and "sneaky throws"), but it looks like that only happened for NATS subscription ([code](https://github.com/Workiva/frugal/blob/3cf3c22ee15636de2cf2790eb5fd968e637429f6/lib/java/src/main/java/com/workiva/frugal/transport/FNatsSubscriberTransport.java#L120-L121)).  For RPC, I think NATS drops the message (client timeout) and HTTP would return 500 (rather than TApplicationException.INTERNAL_ERROR), so I think `FHttpTransport` will convert that to a vague `TTransportException`, which seems like it will conflate "client<->server network error" and "internal service error".

examples/java/src/main/java/examples/ExceptionHandlerServerMiddleware.java
```diff
@@ -0,0 +1,53 @@
...
+                } catch (TTransportException e) {
+                    // See https://github.com/Workiva/frugal/blob/master/lib/java/src/main/java/com/workiva/frugal/exception/TTransportExceptionType.java
+                    // and https://github.com/Workiva/messaging-sdk/blob/master/lib/java/sdk/src/main/java/com/workiva/messaging_sdk/exception/TTransportExceptionType.java
+                    errorCode = e.getType();
```
brettkail-wk, line 38: It seems somewhat confusing to assign this to the same variable since the TApplicationException 1 and TTransportException 1 mean different things (they have different "namespaces").

PR #2330, opened 2023-10-09
FEDX-418 : Update generator.go to remove disposable.Disposable

alishadalal-wk:
Update frugal compiler code to remove disposable.Disposable to fix analyzer issue:

```
error: The private name '_awaitableFutures', defined by 'disposable.Disposable', conflicts with the same name defined by 'Disposable'.
Try removing 'disposable.Disposable' from the 'with' clause.
```

### DIFF
--- a/compiler/generator/dartlang/generator.go
+++ b/compiler/generator/dartlang/generator.go
@@ -2033,17 +2033,9 @@ func (g *Generator) generateClient(service *parser.Service) string {
 
 	// Generate client class
 	if service.Extends != "" {
-		// This generated ignore statement can be removed once we drop support for Dart 2.7.2
-		contents += fmt.Sprint("// The below ignore statement is only needed to workaround https://github.com/dart-lang/sdk/issues/29751, which is fixed on Dart 2.8.0 and later.\n" +
-			"// Dart versions before 2.8.0 need this ignore to analyze properly.\n")
-		contents += fmt.Sprint("// ignore: private_collision_in_mixin_application\n")
 
-		disposableExtensionStr := " with disposable.Disposable"
-		if g.genNullsafe {
-			disposableExtensionStr = ""
-		}
-		contents += fmt.Sprintf("class %s extends %sClient%s implements F%s {\n",
-			clientClassname, g.getServiceExtendsName(service), disposableExtensionStr, servTitle)
+		contents += fmt.Sprintf("class %s extends %sClient implements F%s {\n",
+			clientClassname, g.getServiceExtendsName(service), servTitle)
 	} else {
 		contents += fmt.Sprintf("class %s extends disposable.Disposable implements F%s {\n",
 			clientClassname, servTitle)
--- a/compiler/generator/dartlang/generator.go
+++ b/compiler/generator/dartlang/generator.go
@@ -432,7 +432,7 @@ func (g *Generator) GenerateConstantsContents(constants []*parser.Constant) erro
 	// Add ignores to make lints less noisy in dart consumers
 	ignores := "// ignore_for_file: unused_import\n"
 	ignores += "// ignore_for_file: unused_field\n"
-	
+
 	if _, err = file.WriteString(ignores); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

compiler/generator/dartlang/generator.go
```diff
@@ -2038,8 +2038,12 @@ func (g *Generator) generateClient(service *parser.Service) string {
 			"// Dart versions before 2.8.0 need this ignore to analyze properly.\n")
 		contents += fmt.Sprint("// ignore: private_collision_in_mixin_application\n")
 
-		contents += fmt.Sprintf("class %s extends %sClient with disposable.Disposable implements F%s {\n",
-			clientClassname, g.getServiceExtendsName(service), servTitle)
+		disposableExtensionStr := " with disposable.Disposable"
+		if g.genNullsafe {
```
brettkail-wk, line 2042: Out of curiosity, why is this gated by null-safe?  I don't know enough about Dart, and the error in the PR description doesn't makes sense to me.  Why is this only a problem with that option enabled?

It seems like after this change, we're relying on the superclass to have  `extends disposable.Disposable`.  Does that approach work regardless (i.e., can we remove the `with disposable.Disposable` unconditionally)?

greglittlefield-wf: > Why is this only a problem with that option enabled?

That's a good question. Opting into null safety requires opting into a newer Dart [language version](https://dart.dev/guides/language/evolution#language-versioning) (2.12 specifically), and the language versions sometimes come along with other behavior changes within the analyzer and compiler.

I can't find an exact reference to a change in 2.12 that would only make this start being a problem then, but my guess is that it's an undocumented behavior change or bug.


> It seems like after this change, we're relying on the superclass to have extends disposable.Disposable. Does that approach work regardless (i.e., can we remove the with disposable.Disposable unconditionally)?

I think if we can safely assume that the superclass always `extends disposable.Disposable`, then doing it unconditionally sounds good to me and @robbecker-wf. (Though I suppose we'd also want to validate that assumption even if we did it conditionally). @brettkail-wk do you happen to know if that's a safe assumption?

brettkail-wk: Yes, I believe it is a safe assumption.  The `extends` here is extending another class generated by this same code path: in a `.frugal` file, the `service X extends Y` is just like Dart, so when you generate `XClient` with all the methods in the service, you want to avoid needing to redundantly generate all the methods in `Y` too, so it instead just extends the `YClient` that was generated by the `else`.

(That was confusing to write, so if it doesn't make sense, let me know and I can try again...)

greglittlefield-wf: Okay sweet, yeah that's what it seemed like, but I couldn't tell if that base class could potentially be some other class from somewhere else.

Thank you!!

compiler/generator/dartlang/generator.go
```diff
@@ -2038,8 +2038,12 @@ func (g *Generator) generateClient(service *parser.Service) string {
 			"// Dart versions before 2.8.0 need this ignore to analyze properly.\n")
 		contents += fmt.Sprint("// ignore: private_collision_in_mixin_application\n")
```
robbecker-wf, line 2039: You can remove this comment and ignore line that gets added too

compiler/generator/dartlang/generator.go
robbecker-wf: Great! @alishadalal-wk Can you update this to remove the with disposable.Disposable always? No need to check for nullsafe mode then.

alishadalal-wk: Updated the code for the same.

compiler/generator/dartlang/generator.go
alishadalal-wk: Removed this comment and the ignore line

robbecker-wf (approved):

rmconsole-wf (approved):
+1 from RM

PR #2338, opened 2023-10-12
FEDX-423 : Resolve analyzer error in some repos

alishadalal-wk:
Resolve the following analyzer error in some files of the nullsafe generated code

`The property 'length' can't be unconditionally accessed because the receiver can be 'null'. Try making the access conditional (using '?.') or adding a null check to the target ('!').`

This is present in some of the repos like **content-aggregation-frugal** and **attachments-package-frugal**

### DIFF
--- a/compiler/generator/dartlang/generator.go
+++ b/compiler/generator/dartlang/generator.go
@@ -1257,7 +1257,6 @@ func (g *Generator) generateWrite(s *parser.Struct, kind structKind) string {
 
 		tmpElem := g.GetElem()
 		tmpField := parser.FieldFromType(field.Type, tmpElem)
-		contents += tabtab + fmt.Sprintf("final %s = %s;\n", tmpElem, fName)
 
 		var isSet bool
 		var isNull bool
@@ -1282,6 +1281,11 @@ func (g *Generator) generateWrite(s *parser.Struct, kind structKind) string {
 			isNull = !g.isDartPrimitive(g.Frugal.UnderlyingType(field.Type))
 		}
 
+		if isNull {
+			contents += tabtab + fmt.Sprintf("final %s = %s;\n", tmpElem, fName)
+		} else {
+			contents += tabtab + fmt.Sprintf("final %s = %s%s;\n", tmpElem, fName, g.notNullOperator)
+		}
 		ind := ""
 		if isSet || isNull {
 			ind = tab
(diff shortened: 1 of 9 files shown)


aviary-wf:
## Security Insights

## Action Items
---

compiler/generator/dartlang/generator.go
```diff
@@ -1282,6 +1281,11 @@ func (g *Generator) generateWrite(s *parser.Struct, kind structKind) string {
 			isNull = !g.isDartPrimitive(g.Frugal.UnderlyingType(field.Type))
 		}
 
+		if isNull {
+			contents += tabtab + fmt.Sprintf("final %s = %s;\n", tmpElem, fName)
+		} else {
+			contents += tabtab + fmt.Sprintf("final %s = %s%s;\n", tmpElem, fName, g.notNullOperator)
```
brettkail-wk, line 1287: This is wrong.  `isNull` means "needs an is-null check", and `isSet` means "needs an is-set check" (which might also do a null check).  Even if `isSet` is true (or neither is true), I don't think it's safe to assert non-null because the underlying `writeX` methods are null-safe.

> The property 'length' can't be unconditionally accessed because the receiver can be 'null'. Try making the access conditional (using '?.') or adding a null check to the target ('!').

Can you show the relevant snippet of code?

patkujawa-wf: https://github.com/Workiva/attachment_packager_frugal/pull/129#pullrequestreview-1671467922

patkujawa-wf: @brettkail-wk how does dart nullability interact with <a href="https://github.com/Workiva/attachment_packager_frugal/pull/129#pullrequestreview-1674920883">use_null_for_unset</a>? Seems like that's (part of?) the issue.

brettkail-wk: They're orthogonal/unrelated.  `nullsafe` adds `?` to all fields (because all fields are unsettable in-memory).  `use_null_for_unset` is a generated code optimization and change to semantics (by default, `X()..x = null` is weirdly different from not assigning the field at all).  Ideally, everything would `use_null_for_unset`, so code should continue using that if it already is.

brettkail-wk: So, the scenario is `use_null_for_unset` + `required` + `list` (or map/set).  In that scenario, we currently don't generate an `if (x != null)`, so we do want the `!`, but we don't want it for default required fields.

I've opened an alternative fix in #2339.  (It looks big, but I added new tests/baselines in separate commits.)

alishadalal-wk:
Closing this PR as the changes are carried out [here](https://github.com/Workiva/frugal/pull/2339)

PR #2350, opened 2023-10-23
Add an option that will allow callers to decide if dart sdk annotations are included. 👨‍💼

tomconnell-wf:
### Story:
doc_plat_client is running in unsound null-safety to allow incremental conversion.  This means that it needs to specify which code is not null-safe by putting a `// @dart=2.11` annotation at the top of each unsafe file.  The generator already has the capability to generate this comment, but no control for callers to toggle the ability.

This PR adds an option which, when present, turns on the annotation of output files.

### Acceptance Criteria:

### How To Test:
1. Pull this into doc_plat_client, and run `make frugal`. Your working directory should show no annotations.
2. Add the `adddartcomment` option in the Makefile `frugal-dart` target. (Change lines with `-gen dart:...` to be `-gen dart:adddartcomment...`

### My Test Results:
Generating dpc frugal.  Look at the dart code:  https://github.com/Workiva/doc_plat_client/compare/tom/dart3_frugal_annotations_test?expand=1

#### Reviewers:
@Workiva/service-platform


aviary3-wk:
## Security Insights

## Action Items
---

compiler/generator/dartlang/generator.go
```diff
@@ -47,13 +47,15 @@ const (
 	useInt64              = "use_int64"
 	useVendorOption       = "use_vendor"
 	nullsafe              = "nullsafe"
-	addDartComment        = false // Enable this when debugging Dart generated code
+	addDartCommentOption  = "adddartcomment" // Useful when generating unsound null-safe code (pubspec.yaml sdk >= 2.12, some files annotated at 2.11)
```
robbecker-wf, line 50: Instead of having this constant here you can define a new command line option here https://github.com/Workiva/frugal/blob/master/compiler/generator/generator.go#L81
That location adds it to the help content.

I think the rest is ok with me.

robbecker-wf: or possibly in addition to defining it here, add it in the other place ?

tomconnell-wf: I don't know the import dependencies, I just followed the other options and put in a string.  Thanks for letting me know about this.

compiler/generator/generator.go
```diff
@@ -79,6 +79,7 @@ var Languages = LanguageOptions{
 		"use_null_for_unset": "Use null for unset fields, which disallows null-but-set fields",
 		"use_vendor":         "Use specified import references for vendored includes and do not generate code for them",
 		"nullsafe":           "Generate nullsafe Dart code",
+		"adddartcomment":     "Adds unsound null-safety `// @dart=2.11` comment at the beginning of files, or `// @dart=2.12` if nullsafe option is also present",
```
robbecker-wf, line 82: 👍

brettkail-wk, line 82: Do we need the `@version` for non-nullsafe?  If yes, after the nullsafe transition, at what point would we bump that version?  Even for nullsafe, it seems awkward that if you're on Dart 3, `nullsafe,adddartcomment` will output 2.11.

I wonder if `-gen dart:dart_version_comment=2.11` would be better since that gives control to users for this transition (or future) but without requiring the generator to maintain an ongoing opinion about which version should be output.

robbecker-wf (approved):

compiler/generator/generator.go
robbecker-wf: I don't think you should be able to specify which version to add in the command line as you can choose the wrong one. in combination with the nullsafe arg.

My understanding after talking with Tom is that this additional flag is needed for DPC to be able to convert DPC over to unsound null safety and migrate a package at a time, which still being able to combine DPC into a single package.

Once we're done with the transition we will want to remove both of these arguments.

tomconnell-wf: dpc is going to convert to null-safety incrementally using "unsound null-safety."  This means changing pubspec.yaml to sdk >= 2.12, and annotating all files that have not become null-safe with `// @dart=2.11` (or earlier).  

I'd rather revert this change when on dart 3, although that would wind up with a deprecated noop option.  I don't know of other (good) use-cases, especially for frugal, for overriding the pubspec.yaml sdk, and I'd rather not build infrastructure around generically doing it.

tomconnell-wf:
Rather than cause entropy here, I'll just make a tool in dpc.

PR #2399, opened 2024-02-05
FEDX-548 : Remove nullsafe option in frugal compiler, default to on

alishadalal-wk:
1. Once all frugal generated libraries are regenerated with nullsafety, we can circle back to the frugal compiler and remove the nullsafe command line argument and default the option to always generate nullsafe code. 
2. There are plenty of conditional code paths that consider nullsafe or not that can be cleaned up too.


aviary-wf:
## Security Insights

## Action Items
---

compiler/generator/generator.go
```diff
@@ -79,7 +79,6 @@ var Languages = LanguageOptions{
 		"use_int64":          "Use fixnum.Int64 for i64 fields",
 		"use_null_for_unset": "Use null for unset fields, which disallows null-but-set fields",
 		"use_vendor":         "Use specified import references for vendored includes and do not generate code for them",
-		"nullsafe":           "Generate nullsafe Dart code",
```
brettkail-wk, line 82: This breaks the build of ~all consumers using this option, so I don't think we can remove this.  We can change the default behavior (so that it's always on) and document the option as being deprecated/always on.  Then, after that's been released (and if we think it's worth it?), we could remove the use of the option from all consumers, and then remove the option.

compiler/compiler_test.go
```diff
@@ -75,7 +75,7 @@ func TestValidDartFrugalCompiler(t *testing.T) {
 func TestValidDartNullsafe(t *testing.T) {
 	options := compiler.Options{
 		File:    frugalGenFile,
-		Gen:     "dart:nullsafe",
```
brettkail-wk, line 78: If we're forcing the option to always be on, we can remove this test and the corresponding `dart.nullsafe` files.

alishadalal-wk: Removed the test and `dart.nullsafe` files

compiler/generator/generator.go
alishadalal-wk: Reverted the changes & updated code to generate nullsafe code by default.

compiler/compiler_test.go
```diff
@@ -182,7 +145,7 @@ func TestValidDartUseNullForUnset(t *testing.T) {
 func TestValidDartUseNullForUnsetSafe(t *testing.T) {
 	options := compiler.Options{
 		File:    frugalGenFile,
-		Gen:     "dart:use_null_for_unset,nullsafe",
+		Gen:     "dart:use_null_for_unset",
```
brettkail-wk, line 148: Might as well leave it here to keep coverage of the no-op option?

compiler/generator/generator.go
```diff
@@ -79,7 +79,7 @@ var Languages = LanguageOptions{
...
+		"nullsafe":           "Generate nullsafe Dart code. It is deprecated (by default always on)",
```
brettkail-wk, line 82: This is fine, but perhaps:
> This option has no effect (nullsafe code is always generated).  It remains for backwards compatibility.

compiler/generator/dartlang/generator.go
```diff
@@ -72,19 +72,13 @@ func NewGenerator(options map[string]string) generator.LanguageGenerator {
...
-	if _, ok := options[nullsafe]; ok {
-		generator.genNullsafe = true
-		generator.nullableOperator = "?"
-		generator.notNullOperator = "!"
-		generator.sdkRange = nullsafeDartSdkRange
-		generator.dartComment = "// @dart=2.12"
-		generator.collectionVer = nullsafeCollectionVer
-		generator.thriftVer = nullsafeThriftVer
-	}
+	generator.genNullsafe = true
```
brettkail-wk, line 75: I think we can remove all the options (hardcode the constants) and simplify all the code?

robbecker-wf: I agree. There are places in the code that check these and do conditional logic. Those checks can be removed and just always generate the nullsafe code now.
Bonus points for doing a commit per generator field removal. For instance, 1 commit for removing `generator.nullableOperator = "?"` and replacing uses of it to hard coded `?`.

compiler/compiler_test.go
alishadalal-wk: Re-added the same.

compiler/generator/dartlang/generator.go
alishadalal-wk: Updated the code for same.

robbecker-wf (approved):
+10 verified that regenerating licensing_frugal resulted in no null safety changes to the code (only renaming the temporary elem### variable names since those changed position and is expected)

robbecker-wf:
Oh and verified that the `nullsafe` cmd line option is still present, but does nothing. It does not break when passed or not passed.

robbecker-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #2436, opened 2024-03-28
Java: shutdown executor when a servlet is destroyed

aaronforest-wf:
### Story:
The executors provided to our nats servers are shutdown when the server is stopped, but the executors provided to our http servlets are not shut down when the servlet is destroyed.

Note: I only updated the FJakartaServlet to shutdown its executor service on servlet destroy.  I can apply the same change to the deprecated javax FServlet if it would be valuable.

### Acceptance Criteria:

### Design Notes:
TODO

### How To Test:
TODO

### My Test Results:
TODO

#### Reviewers:
@Workiva/service-platform


aviary3-wk:
## Security Insights

## Action Items
---

lib/java/src/main/java/com/workiva/frugal/server/FJakartaServlet.java
```diff
@@ -112,6 +112,11 @@ private FJakartaServlet(Builder b) {
         this.eventHandler = b.eventHandler != null ? b.eventHandler : new FDefaultServerEventHandler(5000);
     }
 
+    @Override
+    public void destroy() {
+        exec.shutdown();
```
brettkail-wk, line 117: This seems wrong: the servlet didn't create the executor, so it doesn't own it, so it shouldn't be shutting it down (perhaps the executor outlives the servlet for whatever reason).

aaronforest-wf: Would you be okay with optionally shutting down the executor based on a flag set when the servlet is created?  I'm guessing most consumers who provide an executor want it to be shut down when the servlet is destroyed, but I can see how someone might share an executor across multiple servlets and in that case, they wouldn't want it to be shut down.

brettkail-wk: Adding a destroy hook to the servlet seems weird to me.  How are your stopping your processor or its dependencies (database connections, etc.)?  It seems like you should stop the executor in the same way.  i.e., if you're using Spring, define the executor as a bean with a custom destroy method, and inject that as a dependency into your servlet bean.

aaronforest-wf: Yeah, that could work.  I was leaning towards something like this to guarantee that we don't shut down our executor before the servlet is destroyed.  We are using spring though and after reading more, it looks like it might already enforce that beans are destroyed in reverse order of their dependency relationships.

brettkail-wk: Right, it has to shutdown in reverse order or you would have all kinds of issues with the handler itself (i.e., your database connection would be severed while requests were in flight, or before rejecting new requests, etc.).

I don't think it makes sense to merge this PR.

PR #2449, opened 2024-04-11
dart: make element number per file

brianshannan-wf:
### Story:
Currently, the dart generator creates variables using a counter that always increases. Because of this, changing one part of a frugal file can cascade changes to many dart files because that counter creates new variables for all files generated after the file which created more ids than it previously did. It would be nice if this wasn't the case, and only the dart file needing regeneration would produce a diff.

We should be able to track the element counter per file, as these variables are never exposed outside a file, there shouldn't be any conflicts. One potential issue with this approach is if we start writing a file, switch to writing a second file, then come back to the first file, our element counter may now collide. I don't think we ever do this, though.

### Acceptance Criteria:

### Design Notes:
TODO

### How To Test:
- using the dart generator, observe the element counter restarts for each file
- no element collisions are observed in a dart file

### My Test Results:
TODO

### DIFF
--- a/compiler/generator/base.go
+++ b/compiler/generator/base.go
@@ -80,6 +80,10 @@ func (b *BaseGenerator) GetElem() string {
 	return s
 }
 
+func (b *BaseGenerator) ResetElem() {
+	b.elemNum = 0
+}
+
 // ServiceMethodStruct is a service method _args or _result struct.
 type ServiceMethodStruct struct {
 	*parser.Struct
--- a/compiler/testdata/expected/dart.int64/actual_base/f_nested_thing.dart
+++ b/compiler/testdata/expected/dart.int64/actual_base/f_nested_thing.dart
@@ -68,16 +68,16 @@ class nested_thing implements thrift.TBase {
       switch (field.id) {
         case THINGS:
           if (field.type == thrift.TType.LIST) {
-            thrift.TList elem203 = iprot.readListBegin();
-            final elem207 = <t_actual_base_dart.thing>[];
-            for(int elem206 = 0; elem206 < elem203.length; ++elem206) {
-              final elem205 = t_actual_base_dart.thing();
-              t_actual_base_dart.thing elem204 = elem205;
-              elem205.read(iprot);
-              elem207.add(elem204);
+            thrift.TList elem0 = iprot.readListBegin();
+            final elem4 = <t_actual_base_dart.thing>[];
+            for(int elem3 = 0; elem3 < elem0.length; ++elem3) {
+              final elem2 = t_actual_base_dart.thing();
+              t_actual_base_dart.thing elem1 = elem2;
+              elem2.read(iprot);
+              elem4.add(elem1);
             }
             iprot.readListEnd();
-            this.things = elem207;
+            this.things = elem4;
           } else {
             thrift.TProtocolUtil.skip(iprot, field.type);
           }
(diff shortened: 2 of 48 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

compiler/testdata/expected/dart.int64/actual_base/f_nested_thing.dart
```diff
@@ -98,12 +98,12 @@ class nested_thing implements thrift.TBase {
     validate();
 
     oprot.writeStructBegin(_STRUCT_DESC);
-    final elem208 = things;
-    if (elem208 != null) {
+    final elem5 = things;
```
brettkail-wk, line 101: I guess the counter could be reset per method to further reduce noise.

brettkail-wk (approved):

compiler/testdata/expected/dart.int64/actual_base/f_nested_thing.dart
brianshannan-wf: I debated doing that, I decided to do this instead because it would be a fair amount simpler
- GenerateFile can't recursively call itself, unlike some other generator methods, so we don't have to worry about resetting when we already have some generated variables in scope
- there's only one place we need to reset with this, as opposed to a few different method generation functions
and I think this is a large enough improvement as is.

brettkail-wk: > GenerateFile can't recursively call itself, unlike some other generator methods

Sorry, I meant per _generated_ method (e.g., in `generateRead`), not the top of each `Generator` Go method (e.g., in `generateReadFieldRec`).

> there's only one place we need to reset with this, as opposed to a few different method generation functions

Agreed, I already approved and was just thinking out loud.

jacobfeddersen-wf (approved):

compiler/generator/base.go
```diff
@@ -80,6 +80,10 @@ func (b *BaseGenerator) GetElem() string {
 	return s
 }
 
+func (b *BaseGenerator) ResetElem() {
+	b.elemNum = 0
+}
```
adamdockter-wk, line 85: I assume two things based on this and just want to verify they are correct:
1. there is a single b *BaseGenerator instance being used for generation?
2. The GenerateFile method is never call in parallel, because this counter is shared state in the single generator?

I think those assumption have to be true to make this work?  If so could you comment them so we don't accidentally violate them in the future.

brianshannan-wf: Both of those are true
1. Each generator embeds a BaseGenerator. I'm not sure how multiple BaseGenerators would work, but doing that would reset the element counter anyway
2. None of the generator code is threadsafe, so calling in parallel would be a problem now. I think the thing to be more concerned about is GenerateFile being recursively called by itself so the element counter could be reset and re-incremented when we go back to the original file. I could not find a way for this to happen.

I can add a comment with that assumption

brettkail-wk (approved):

adamdockter-wk (approved):

jacobfeddersen-wf:
QA+1
- [ ] Dart files have elem numbers generated per-file
- [ ] Elems do not conflict and consuming this version of frugal still generates working code

brianshannan-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

